PR #22, opened 2019-06-06
docs: add initial recommendations for testing

wincent:
Just a tiny recommendation to seed the document. There are lots more things that could be said about testing, and it can get pretty subjective, so I thought I'd start with something relatively uncontroversial, but which has come up in code review quite a bit.


wincent:
For another in-flight discussion about testing, see: https://github.com/liferay/liferay-npm-tools/issues/108

PR #28, opened 2019-06-11
docs: add Aspect Oriented Programming document

carloslancha:


guidelines/dxp/metal-plugins/metal-aop.md
```diff
@@ -0,0 +1,70 @@
...
+Once dependency is added AOP shall be imported in the js module:
+
+`import {AOP} from 'frontend-js-web';`
+
+## Use case for AOP.after
+
+Example found in `layout-content-page-editor-web/src/main/resources/META-INF/resources/js/utils/FragmentsEditorDragDrop.es.js`
+
+```
+_getRegionInterceptor =
```
jbalsas: We can probably design a simpler use case to highlight this. For example, in this case, it would be something like:

// Use case: Change the return value of an instance method invocation
```
import {AOP} from `frontend-js-web`;

const foo = new Foo();
console.log(foo.bar()); // "bar"

const _interceptedFoo = AOP.after(target => {
        return AOP.alterReturn("modified bar")
    }, foo, 'bar');

console.log(foo.bar()); // "modified bar"

File: guidelines/dxp/metal-plugins/metal-aop.md [hunk 2]
```diff
@@ -0,0 +1,70 @@
+# Metal AOP guide of use
```
jbalsas: I'm not sure how to document this, but we should avoid `metal` references... 

My original idea was something like `/patterns/dxp/aspect_oriented_programming.md` or if `AOP` is not a clear enough reference, something named after the use cases...

jbalsas: @wincent, @julien, @bryceosterhaus, input on this?

carloslancha:
@jbalsas done!

patterns/dxp/aspect_oriented_programming.md
```diff
@@ -0,0 +1,53 @@
+# Aspect Oriented Programming guide of use
```
wincent: Suggest:

```
# Aspect Oriented Programming (AOP) guidelines
```

ie. 
- introduce "AOP" acronym before first usage below
- "guidelines" is *more betterer* English than "guide of use"

patterns/dxp/aspect_oriented_programming.md
```diff
@@ -0,0 +1,53 @@
...
+
+> **NOTE:** This plugin was originally in metal-plugins but now lives in portal so we can provide a better developing experience.
+
+AOP provides tools to improve control on what is executed before and after other method is invoked. You can finde this component in `frontend-js-web` module in portal.
```
wincent: - "control on" → "control of"
- "other method" → "another method"
- "finde" → "find"
- "`frontend-js-web` module" → "the `frontend-js-web` module" (bonus points if you link it, in which case you can delete the "in portal" bit.

patterns/dxp/aspect_oriented_programming.md
```diff
@@ -0,0 +1,53 @@
+# Aspect Oriented Programming guide of use
+
+> **NOTE:** This plugin was originally in metal-plugins but now lives in portal so we can provide a better developing experience.
```
wincent: I've been calling it "liferay-portal" in the other docs to match the repo name on GitHub, and making it a link.

patterns/dxp/aspect_oriented_programming.md
```diff
@@ -0,0 +1,53 @@
...
+
+AOP provides tools to improve control on what is executed before and after other method is invoked. You can finde this component in `frontend-js-web` module in portal.
+
+See more about Aspect Oriented Programming in [https://medium.com/@kyuwoo.choi/sneak-peek-to-javascript-aop-16458f807842](this article)
```
wincent: Missing a period at the end of the sentence here.

patterns/dxp/aspect_oriented_programming.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Preparing the use
```
wincent: "the use" sounds a bit weird. I'd just call this one:

```
## Prerequisites
```

patterns/dxp/aspect_oriented_programming.md
```diff
@@ -0,0 +1,53 @@
...
+
+First of all the `frontend-js-web` dependency shall be added to the `package.json` of the module in which we'll use AOP. Be sure that the version matches with the one in `frontend-js-web/bnd.bnd` file.
```
wincent: - "shall" → "should"
- "matches with" → "matches"
- "`frontend-js-web/bnd.bnd` file" → "`frontend-js-web/bnd.bnd`" (again, could link here).

patterns/dxp/aspect_oriented_programming.md
```diff
@@ -0,0 +1,53 @@
...
+"dependencies": {
+	...
+	"frontend-js-web": "4.0.0",
+	...
+},
+```
+
+## Use case: Change the return value of an instance method invocation
+
+```
```
wincent:
```
```javascript
```

patterns/dxp/aspect_oriented_programming.md
```diff
@@ -0,0 +1,53 @@
...
+import {AOP} from `frontend-js-web`;
+
+const foo = new Foo();
+console.log(foo.bar()); // "bar"
```
wincent: I'd leave out the `console.log()` in the examples as they increase the clutter. (Feel free to keep them though if you think it's better.)

patterns/dxp/aspect_oriented_programming.md
```diff
@@ -0,0 +1,53 @@
...
+const _interceptedFoo = AOP.before(target => {
+		return AOP.prevent();
+	}, foo, 'bar');
+
+console.log(foo.bar()); // undefined
+```
+
+## Use case in portal
+
+You can find an example of use of AOP in https://github.com/liferay/liferay-portal/blob/master/modules/apps/layout/layout-content-page-editor-web/src/main/resources/META-INF/resources/js/utils/FragmentsEditorDragDrop.es.js
```
wincent: Can we make this a bit prettier?

```
in [FragmentsEditorDragDrop.es.js](...)
```

Also, because files move around, we should make an "evergreen" link that won't break (you can get this by hitting "y" on GitHub):

```
https://github.com/liferay/liferay-portal/blob/e6260e573596c922696b2bf66ece45f1ca32d733/modules/apps/layout/layout-content-page-editor-web/src/main/resources/META-INF/resources/js/utils/FragmentsEditorDragDrop.es.js
```

carloslancha:
@wincent done! Thx!

patterns/dxp/aspect_oriented_programming.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Use case: Change the return value of an instance method invocation
+
+```javascript
+import {AOP} from `frontend-js-web`;
+
+const foo = new Foo();
+foo.bar(); // "bar"
+
+const _interceptedFoo = AOP.after(target => {
```
jbalsas: We have an unused const here which is a bit fishy. What's this return value and what can you do with it?

We should probably want to document how the handle contains a `detach` method to clean up when the interception is no longer required (e.g. during the component destroy lifecycle)

patterns/dxp/aspect_oriented_programming.md
bryceosterhaus: liferay-portal also helps distinguish from react portals

patterns/dxp/aspect_oriented_programming.md
carloslancha: Done!

PR #29, opened 2019-06-12
docs: add initial information about requests

julien:


patterns/dxp/requests.md
```diff
@@ -0,0 +1,35 @@
+# Requests
+
+Making requests is a common task in the frontend world.
+
+In [`liferay-portal`](https://github.com/liferay/liferay-portal/), we provide the [`fetch`](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/liferay/util/fetch.es.js) function which is a very thin wrapper on top of the existing [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) function implemented in modern browsers. Additionally, a [polyfill](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-compatibility/frontend-compatibility-ie/build.gradle#L351) is also served for Internet Explorer 11, to ensure the `fetch` function is defined.
```
wincent: I suggest hitting "y" on GitHub when you're grabbing links like:

https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/liferay/util/fetch.es.js

That turns it into an "evergreen" URL that will never break unless we rewrite history (which we won't):

https://github.com/liferay/liferay-portal/blob/73fa9b927147e5a547a2cce44dd1b46ae9ab390d/modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/liferay/util/fetch.es.js

It's all too easy for these links to become dead and broken at the next rename, and that's worse than looking at something slightly stale, I think.

This is especially the case in your second link, which points at a specific line number: that kind of link will easily break and start pointing at the wrong line unless you grab one pointing at a specific commit by hitting "y".

patterns/dxp/requests.md
```diff
@@ -0,0 +1,35 @@
...
+fetch(url)
+	.then(res => res.json())
+	.then(res => {
+		// Use the data returned by the request
+	})
+```
+
+## More information.
+
+Please visit the MDN documentation on the [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) A.P.I. for more information.
```
wincent: Probably just "API" here. You didn't write M.D.N., after all. 😂

julien: Thanks, will update

jbalsas:
We should probably mention that the fetch wrapper is accessed as Liferay.Util.fetch from those places where importing is not possible

patterns/dxp/requests.md
```diff
@@ -0,0 +1,35 @@
...
+	...
+	"frontend-js-web": "4.0.0",
+	...
+},
+```
+
+### Use case: making a request to and endpoint that returns `json`
+
+```
+import {fetch} from `frontend-js-web`;
```
jbalsas: I think we talked about this in Clay and actually agreed that backticks for everything didn't make that much sense, right?

If so, I wouldn't expect to see it in our code samples.

patterns/dxp/requests.md
```diff
@@ -0,0 +1,35 @@
...
+
+const url = `http://localhost:8080/path/to/an/existing/endpoint`;
```
jbalsas: Single quote?

patterns/dxp/requests.md
```diff
@@ -0,0 +1,35 @@
...
+
+fetch(url)
+	.then(res => res.json())
+	.then(res => {
+		// Use the data returned by the request
+	})
```
jbalsas: missing semicolon? 😝

patterns/dxp/requests.md
julien: Yes, sorry that's my mistake.

PR #32, opened 2019-06-13
docs: fix footer reference on dev_dependencies.md

rdai10:
I noticed that the footnote doesn't properly link to the `Related resources` section.


wincent:
Thanks @rdai10.

PR #59, opened 2019-07-09
docs: add dxp/debouncing.md

carloslancha:


dxp/debouncing.md
```diff
@@ -0,0 +1,17 @@
+# Debouncing guidelines
+
+> **NOTE:** This plugin was originally in metal-plugins but now lives in [liferay-portal](https://github.com/liferay/liferay-portal) so we can provide a better developing experience.
+
+See more about debouncing in [this article](https://davidwalsh.name/javascript-debounce-function).
```
wincent: At the risk of being pedantic, I don't think the example (a resize listener) given in the article is a very good one. For that use case I think you'd want throttle, not debounce (scroll listeners are another example where throttle is better than debounce). In those use cases you don't want to delay providing feedback to the user until the interaction is over (the interaction could last indefinitely); rather, you want to slow the rate of updates, so throttle is what you want. A good example use case for debounce is a type-ahead/autocomplete field, because ongoing interaction is likely to invalidate any work you start doing during the interaction, so you only want to do it once the user has paused long enough to make it meaningful.

In other words, I think it's fine to link to the article, but I would add some guidelines on when to use debounce and when to use throttle.

dxp/debouncing.md
```diff
@@ -0,0 +1,17 @@
+# Debouncing guidelines
+
+> **NOTE:** This plugin was originally in metal-plugins but now lives in [liferay-portal](https://github.com/liferay/liferay-portal) so we can provide a better developing experience.
```
wincent: Given what @jbalsas [said about avoiding Metal references](https://github.com/liferay/liferay-frontend-guidelines/pull/28#discussion_r292521996), maybe we don't need this sentence.

dxp/debouncing.md
jbalsas: Sounds pedantic and informative at the same time :)

dxp/debouncing.md
jbalsas: Yeah, maybe we can simply gaslight the world into thinking this never happened :D

carloslancha: I just copied it from AOP guidelines. I'll remove it from both then.

carloslancha:
@wincent @jbalsas done!

PR #62, opened 2019-07-16
docs: add recommendations for testing UI components

wincent:
As evidenced by this mega thread:

https://github.com/jbalsas/liferay-portal/pull/1819#discussion_r303149958

This is a very complex topic with lots of tradeoffs and historical context to weigh up. As such, this PR seeks to make some modest progress by starting with some minimal guidelines. We begin with the bits that are clear and uncontroversial. As we move deeper into the topic, things become more complicated and harder to make recommendations on (and we probably get diminishing returns anyway).

### DIFF
--- a/general/testing.md
+++ b/general/testing.md
@@ -1,5 +1,19 @@
 # Recommendations for testing
 
+## Testing UI components
+
+-   Snapshots SHOULD only be used as "smoke tests": they are very good for verifying that a component renders at all, but they are _not_ good for communicating intent or what is important. For example, when something in a large snapshot changes, there is nothing intrinsic in the snapshot to communicate whether any particular part of it is important.
+-   Shipping test attributes to the client is overtly bad because it increases network payloads, parse times, DOM bloat etc, and ends up becoming an unofficial API.
+    -   Therefore, we MUST NOT send "data-testid" to the client.
+-   The code and markup that is exercised in the test environment should match that which is used in the production environment, otherwise you are not actually testing the thing that you would like to test.
+    -   Because of the recommendation above, any usage of "data-testid" MUST be stripped from when outside the test environment.
+    -   But because of the desire to not have divergent code paths, "data-testid" SHOULD NOT be used, except as an interim measure.
+-   In the absence of "data-testid", we seek the most stable way to target elements in tests ("stable" in the sense that tests are less likely to break as a result of superficial changes); we think the methods that are most likely to remain stable are the following and therefore recommend that we SHOULD use:
+    -   Attributes or content that are significant to the user (for example, the text of a button such as "Publish"); at the end of the day, the visible (or audible) content and behavior of a product is literally the reason we build products, so the highest value tests should focus on aspects that are perceptible and observable to the user. In the button example, if the text changes, then we _want_ the test to break.
+    -   `id` attributes when a semantically meaningful and unique "identity" can be assigned to an element (obvious considerations about multiple instances in a document needing to be unique apply).
+    -   `class` based selectors where the name is semantically meaningful (ie. describing the role of the element as opposed to controlling the presentational aspects of it).
+    -   other selectors, with a strong preference for _semantic_ elements because these are less likely to change over time (that is a list `<ul>` is likely to remain a list, and a `<nav>` element is likely to continue being for navigation, totally unlike arbitrary container elements like `<span>` or `<div>` that could change at any time).
+
 ## Start `it()` descriptions with a verb, not with "should"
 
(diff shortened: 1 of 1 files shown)


general/testing.md
```diff
@@ -1,5 +1,19 @@
...
+    -   Because of the recommendation above, any usage of "data-testid" MUST be stripped from when outside the test environment.
+    -   But because of the desire to not have divergent code paths, "data-testid" SHOULD NOT be used, except as an interim measure.
+-   In the absence of "data-testid", we seek the most stable way to target elements in tests ("stable" in the sense that tests are less likely to break as a result of superficial changes); we think the methods that are most likely to remain stable are the following and therefore recommend that we SHOULD use:
+    -   Attributes or content that are significant to the user (for example, the text of a button such as "Publish"); at the end of the day, the visible (or audible) content and behavior of a product is literally the reason we build products, so the highest value tests should focus on aspects that are perceptible and observable to the user. In the button example, if the text changes, then we _want_ the test to break.
```
jbalsas: How does that deal with localization?

general/testing.md
```diff
@@ -1,5 +1,19 @@
 # Recommendations for testing
 
+## Testing UI components
+
+-   Snapshots SHOULD only be used as "smoke tests": they are very good for verifying that a component renders at all, but they are _not_ good for communicating intent or what is important. For example, when something in a large snapshot changes, there is nothing intrinsic in the snapshot to communicate whether any particular part of it is important.
+-   Shipping test attributes to the client is overtly bad because it increases network payloads, parse times, DOM bloat etc, and ends up becoming an unofficial API.
+    -   Therefore, we MUST NOT send "data-testid" to the client.
+-   The code and markup that is exercised in the test environment should match that which is used in the production environment, otherwise you are not actually testing the thing that you would like to test.
```
jbalsas: This was something that happened in the past, but I don't have a clear feeling of wether it can be changed or not... if it could be changed, how important do you think this is, shipping and testing the same exact code?

After all, we're doing all sorts of transformations already...

general/testing.md
wincent: I think that in the same way you need to write tests that don't produce different output (and break) when run in different time zones, we need to isolate our test environment from locale differences. That is, I expect almost 100% of test runs to be locked to a specific localization, and then have a small number of instances where we effectively "sample" other localizations to make sure that in principal, things still work.

general/testing.md
wincent: I'd say:

    don't ship test junk to client > ship exactly the same bytes to client

ie. desirable, but not essential, which is why it appears further down the list.

This is going to be subtle, but another thing about "data-testid" though that makes it seem like a potential code smell to me is this:

A "data-testid" says "this thing is important to me". But if it is important, why isn't it targetable in some user-visible way? That is, users care about behavior, content, and behavior of specific things (buttons, sliders, charts etc). If the only way I can label something as important is via *a mechanism that is invisible to the user* (not even in the source code on the client), then there is a risk that I might be testing an implementation detail and not something which a user actually cares about. And if QA folks start leveraging these "data-testid" attributes as a defacto API, then we wind up in a situation where QA may be asserting things at the layer of implementation details, and which isn't the layer that users interact with and care about.

Or maybe another way to say that is that ideally these should all be aligned:

    what users care about === what you actually test === the API you use to test

with the obvious observation that the definition of "API" can include markup etc, not just programming interfaces.

jbalsas: Sounds good to me, just trying to present alternate points for discussion :)

I think it would be good to document this somewhere as rationale. This reasoning carries more weight than "historical reasons prevent us from..."

general/testing.md
jbalsas: Yeah, you're probably right... we don't really have a way to serve the right translations anyways...

One thing that @bfarache brought up, though, was how could we test that different translations reflected properly in the UI (german for example being usually longer than english). I'm thinking this should probably be tested somewhere else, just to double-check.

wincent: > I'm thinking this should probably be tested somewhere else

Yeah, I don't know if we have any tooling that could help us check for catastrophic layout failures (or even non-catastrophic but still unpleasantly usability problems) in an automated way. It seems like it would be pretty brittle even if we had it (how would you distinguish between "broken" layout and merely "different" layout?). Maybe we already do, but part of the manual QA for every new feature could be to test in English, German (because of the long words), and probably a RTL language as well.

And I think we'd want some non-visual tests that run in a language other than English just to minimally exercise the localization machinery (or part of it, because as you said in another thread, the `Liferay.Language` stuff is hard to test end-to-end because it spans so many levels).

jbalsas:
Merging this with the changes, anyone is free to comment and send followup comments :)

PR #63, opened 2019-07-22
docs: shard testing content out into multiple files

wincent:
As requested [here](https://issues.liferay.com/browse/LPS-97586?focusedCommentId=1901126&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1901126), this PR splits out our testing documentation into multiple files and expands it a bit.

> could we try to expand our information ... to maybe cover:
>
> - Accepted libraries
> - Anatomy of a test
> - Examples
>
> Maybe we should split `testing.md` into different linked sections with a generic TOC to aggregate everything?

This PR is just a start — there is still a lot more that could be said — but it gets the ball rolling.


general/testing/README.md
```diff
@@ -0,0 +1,183 @@
...
+>
+> _Interaction/mock testing mindset_
+>
+> Perform an action on some piece of code under test, then verify that internally, some element communicated with some element using an expected value.
+
+### Write tests for the reader, not the writer
+
+In software development, we try to balance a number of nebulous concepts like "readability", "maintainability", and "ergonomics". This in turn leads to maxims like "Don't Repeat Yourself" (DRY).
+
+> In test code, it is **even more important that code be readable and intention-revealing**. Don't worry about duplication in tests: it may actually be a good thing.
```
jbalsas: 😍

general/testing/README.md
```diff
@@ -0,0 +1,183 @@
...
+
+Think from the perspective of the person that is looking at a test failure in CI, in code that they didn't write and may not understand. They _should not have to reverse-engineer the test_ to figure out what the code is supposed to be doing. Instead, they should be able to read it from top-to-bottom and see it all spelled out in a very clear, pedestrian way, without having to jump through a list of calls to helper functions.
+
+> In tests, **dumb is the new clever**.
+
+### Concrete recommendations
+
+#### Assert (ie. `expect()`) one thing per `it()` block
```
jbalsas: Should we enforce this? I'm assuming we'd rather leave it as a recommendation only as we'll probably need to deviate from this...

general/testing/README.md
```diff
@@ -0,0 +1,183 @@
...
+});
+
+describe('positioning controls in the centre', () => {
+	it('has a position property of "center"', () => {
+		expect(getControl('center').position).toBe('center');
+	});
+});
+```
+
+This (contrived) example is one where your intuition may be to "DRY" up the code, but consider the position of the person who reads a report of a "test failure on line 123" and now has to figure out which iteration of the loop. It's better for them if we spell out each example alone.
```
jbalsas: nit: maybe add a link to the DRY section?

general/testing/README.md
wincent: Yeah, that's my thinking as well. When I originally wrote this I even started the sentrence with "Try to..."

This kind of advice is like "try to keep your functions shorter than X lines" — it's good to bear in mind the benefits of short functions, but if you slavishly chop *all* of your code up into bite-sized pieces then sometimes you're going to actually make it worse. So it's good to leave some things up to individual judgment.

general/testing/README.md
wincent: Something like this?

https://github.com/liferay/liferay-frontend-guidelines/pull/63/commits/1f13ba3321ed73b9bf1e2e1bb80545d0fd273118

I'll merge this with that, and if I misunderstood what you meant please feel free to edit.

general/testing/README.md
bryceosterhaus: Im not a huge fan of this rule. Typically because I like to assert before and after whatever action I take. For example if I expect something to happen when I click a button, I typically assert it didnt happen before the click and then assert it did happen after the click.

wincent: @bryceosterhaus: It's not a "rule" though. We can soften the language to make it clear that it is not "one-assertion-or-die" (please feel free to do that, or to add some sentence(s) explaining your use case).

A pre-and-post assertion is obviously quite a sensible thing to do; in fact you could argue that your pre-and-post `expect` calls are actually just two halves of a *single assertion* about a transition between two states. RSpec (a Ruby BDD library) used to have matchers explicitly tailored for this style of the form:

```ruby
expect{Counter.increment}.to change{Counter.count}.from(0).to(1)

expect{Counter.increment}.to change{Counter.count}.by(2)
```

Would be pretty easy to make a JS/Jest equivalent, I'd think:

```javascript
expect(() => button.click()).toChange({
  value: () => component.counter,
  from: 0,
  to: 1,
});

expect(() => button.click()).toChange({
  value: () => component.counter,
  by: 2,
});

// or something else equivalent; this is just one sketch...
```

The main goal of this recommendation is to avoid the kind of insanity in the counter-example in the docs where you have a test that doesn't clearly specify the behavior at all because it is asserting half-a-dozen things at once.

bryceosterhaus: Sounds good to me. I figured that is what you were getting at but just wanted to make sure and clarify.

PR #66, opened 2019-08-09
docs: add documentation concerning URL

julien:


dxp/urls.md
```diff
@@ -0,0 +1,113 @@
...
+### Use case: appending a parameter in a URL
+
+```
+const myUrl = new URL('http://localhost:8080/?foo=baz');
+
+myUrl.searchParams.append('foo', 'another-baz');
+
+console.log(myUrl.toString());
+
+// Will print: http://localhost:8080/?foo=baz&foo=another-baz
```
jbalsas: Maybe we could ellaborate a bit more in which case one would want more than one `foo` param so the use case is clear and not just the technical aspect

dxp/urls.md
```diff
@@ -0,0 +1,119 @@
...
+
+Make sure you read the documentation referenced above to make sure you fully understand the URL API.
+
+Additionally, a polyfill is also served for Internet Explorer 11, to ensure the `URL` API can be used correcly.
+
+
+### Use case: creating a relative URL
+
+```
+
```
wincent: Kill this unnecessary blank line.

dxp/urls.md
```diff
@@ -0,0 +1,119 @@
...
+```
+const myUrl = new URL('http://localhost:8080/?foo=baz&bar=qux');
+
+console.log(myUrl.searchParams.has('foo'));
+console.log(myUrl.searchParams.has('bar'));
+
+// Will print: true (twice)
+
+
```
wincent: Kill this blank line. Or perhaps better yet, drop the "will print" comments and do this instead:

```javascript
console.log(myUrl.searchParams.has('foo')); // true
console.log(myUrl.searchParams.has('bar')); // true
console.log(myUrl.searchParams.has('not-there')); // false
```

dxp/urls.md
```diff
@@ -0,0 +1,119 @@
...
+We now recommend using the native [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) API.
+
+Make sure you read the documentation referenced above to make sure you fully understand the URL API.
+
+Additionally, a polyfill is also served for Internet Explorer 11, to ensure the `URL` API can be used correcly.
+
+
+### Use case: creating a relative URL
+
+```
```
wincent: Can you please make all of these fenced code blocks start with?

```
```javascript
```

dxp/urls.md
```diff
@@ -0,0 +1,119 @@
...
+// Will print: 'baz'
+
+console.log(myUrl.searchParams.get('bar'));
+
+// Will print: 'qux'
+
+console.log(myUrl.searchParams.get('not-there'));
+
+// Will print: null
+```
```
wincent: These also might be better briefer.

```javascript
console.log(myUrl.searchParams.get('foo')); // 'baz'
console.log(myUrl.searchParams.get('bar')); // 'qux'
console.log(myUrl.searchParams.get('not-there')); // null
```

or even?

```javascript
myUrl.searchParams.get('foo'); // 'baz'
myUrl.searchParams.get('bar'); // 'qux'
myUrl.searchParams.get('not-there'); // null
```

"Less is more", and all that. 😂

matuzalemsteles:
Ah my concern in documenting these things is that it seems that we are describing the MDN documentation or duplicating it in some way... Maybe it would be better if it were to compare with `metal-uri` and already link to MDN since it will always be up to date when Something to change.

wincent:
> My concern in documenting these things is that it seems that we are describing the MDN documentation or duplicating it in some way

This is a good point, but:

- The work is already done.
- We don't have to worry about getting "out of sync" with what's on MDN (because there is *no way* the `URL` API is ever going to have breaking changes).
- This document is pretty readable thanks to the headings (compared with the MDN page which just has one big block under "Usage Notes"), so that's useful too.
- We have links to both `metal-uri` (and I wouldn't want to go into more detail about a legacy API, so the link is probably enough) and the MDN pages here.

Which is all a long way of saying I think this change is probably OK.

Let's merge it and anybody reading this thread can feel free to make improvements in another PR.

PR #70, opened 2019-08-16
feat: configure Semantic Pull Request bot

wincent:
Seeing as I just noticed that it's actually been configured for this repo.

### DIFF
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -1,0 +1,7 @@
+---
+# Configuration for the Semantic Pull Request bot.
+# https://github.com/probot/semantic-pull-requests
+
+allowMergeCommits: true
+
+titleAndCommits: true


wincent:
Going to do some tests to see if I can trigger the bot...

wincent:
So it's [supposed to complain](https://github.com/probot/semantic-pull-requests/blob/f004bc492b6d2c43d8bbb64220fae51499b70f6e/lib/is-semantic-message.js#L18) about a bogus type. Not sure if it has to be committed to the repo in order for it to work.

wincent:
OK, well the bot seems to accept anything 😂. I am going to fix this, merge it and then repeat the tests, to see whether it works once the config is on the "master" branch (which I guess makes sense).

PR #76, opened 2019-08-30
docs: document "throttle()"

wincent:
Initial documentation for `throttle()`.

Changes:

- Add info about `throttle()`.
- Remove link to article because it is a bit misleading (it contains a recommendation to use `debounce()` on resize listeners, for instance).
- Remove bit about depending on a specific version: we are about to switch to using "*" versions for internal dependencies.
- Show how to `import` and call.

Related: https://issues.liferay.com/browse/LPS-100749

Possible follow-ups that we might want to do:

- In the context of React and hooks, we may want to show how to use `useCallback()` to preserve referential equality across renders.


jbalsas:
Looks great to me! 👍 

Do you want to wait until `throttle` gets merged to get this in? Or should we do it already?

wincent:
Force pushed because I realize I never explicitly defined what `throttle()` is; added this sentence:

> Throttling ensures that a function not be called more often than every `interval` milliseconds.

I'd be fine with merging it now, as `throttle` should be in soon enough (and the promised change to use `"*"` version ranges too).

PR #83, opened 2019-09-06
docs: css-when-to-use-utility-classes

eduardoallegrini:


css/how_to_name_the_files.md
```diff
@@ -0,0 +1,115 @@
...
+
+Sometimes we need a composed name for our files, example: \_text-editor.scss, \_button-style.scss, \_card-image.scss
+
+The convention defined by Clay and Bootstrap is to use a single `-` hyphen character to create a separation between the words, lowercase Unicode characters to write the names, and the `_` underscore character mentioned in the previous section
+
+To simplify the syncronization process between the frameworks we decided to extend the same pattern to all our projects
+
+So, don't use:
+
+-   \_text_editor.scss
```
jbalsas: We do exactly the opposite in `liferay-portal`, because mixing `_` and `-` is not allowed. We need to either recommend this other approach (and just not follow it in Clay) or make a clear note about this.

eduardoallegrini: Is there a particular reason why mixing _ and - is not allowed?

jbalsas: Brian doesn't like mixing, that's all. Module names have `-` and file names have `_`.

That doesn't mean we have to change this in places like Clay, but we need to note that in DXP, the convention for filenames is different...

PR #86, opened 2019-09-11
docs: css-how_to_name_the_classes

eduardoallegrini:


css/how_to_name_the_classes.md
```diff
@@ -0,0 +1,118 @@
+# How to name the classes?
```
wincent: Can we put "CSS" in the title?

Either as a prefix ("CSS: ") or inline ("CSS classes")

css/how_to_name_the_classes.md
```diff
@@ -0,0 +1,118 @@
...
+
+To understand which is the best name for a css class
```
wincent: Can we use "CSS" in uppercase consistently throughout?

css/how_to_name_the_classes.md
```diff
@@ -0,0 +1,118 @@
...
+
+Creating your code using classes makes your project modular (or modular-ready), it will be easier for you and other developers to export it into multiple parts of your project or even into multiple projects
+
+## 2. The right name
+
+To choose the right name for a class we need to take into consideration the following suggestions
+
+### A. Allowed characters
+
+To avoid undesired errors, use only lowercase Unicode characters for the words and a single `-` hyphen character between each of them, in order to improve the language accessibility, we recommend to use english words.
```
wincent: "English" must always start with a capital "E".

css/how_to_name_the_classes.md
```diff
@@ -0,0 +1,118 @@
...
+.fjord-text-2 {...}
+.fjord-text-3 {...}
+.fjord-text-4 {...}
+```
+
+### D. Component instances
+
+We always recommend the solution proposed in the previous section, but sometimes it will be necessary to change the standard method
+
+> Imagine we need to create 2 images in 2 different components for the Fjord theme, but the differences between these elements are so uniques that we can't afford to generate the required modifiers because they won't be used anywhere else.
```
wincent: uniques -> unique

css/how_to_name_the_classes.md
```diff
@@ -0,0 +1,118 @@
...
+
+In that case we are going to rely on the Atomic Design concepts to add the desired specifications to the class.
```
wincent: Would be a good place to make the words "Atomic Design" a link to a resource on Atomic Design.

css/how_to_name_the_classes.md
```diff
@@ -0,0 +1,118 @@
...
+
+```scss
+// Class instance SCSS
+.fjord-card-image {...}
+.fjord-list-image {...}
+```
+
+### E. Responsive particles
+
+There is a special part of the name is worth mentioning even if we created a dedicated session for this topic: the "responsive particles". [future_link]
```
wincent: "session": do you mean "section"?

css/how_to_name_the_classes.md
```diff
@@ -0,0 +1,118 @@
...
+
+The "responsive particle" is a part of the class that can be `sm`, `md`, `lg`, or `xl` (small, medium, large, or extra-large), and it goes between the `class-property-word` and the `class-value-word`
+
+```scss
+// SCSS example
+.col-sm-6 // create a column that occupy 6/12 of the available container width on devices with a screen width of `576px` or above
```
wincent: occupy -> occupies

Also on the next line.

css/how_to_name_the_classes.md
```diff
@@ -0,0 +1,118 @@
...
+
+## 3. Good practices
+
+### A. RTL ready
+
+To improve the accessibility of the right-to-left languages we should avoid the words `left` and `right` and use `before`, `after`, `prepend`, `append`, `first`, `last`, `previous`, `next`, `back`, `forward` etc
+
+### B. HTML selectors
+
+Instead of using HTML selectors like `.nav > a` you should prefer defined classes like `.nav-link` the result will be lighter and faster for the browser, it will also limit the styles from _bleeding_ to the children tags
```
wincent: Needs a full stop after `.nav-link`. "The result" starts a new sentence. Needs an "and" before "it will also" to avoid a run-on sentence.

css/how_to_name_the_classes.md
jbalsas: :nerd

css/how_to_name_the_classes.md
```diff
@@ -0,0 +1,118 @@
...
+We should create our responsive classes following the same pattern
+
+```scss
+// SCSS example
+.fjord-title-sm-1 {
+	font-size: 20px;
+}
+.fjord-title-md-2 {
+	font-size: 30px;
+}
```
jbalsas: Should these examples be, in fact, responsive? I was expecting to see some media-queries in here to see how `-md-` and `-sm-` would affect different devices. Without it, this seems more or less arbitrary.

PR #87, opened 2019-09-16
docs: css_how-to-write-responsive-css

eduardoallegrini:


css/how_to_write_responsive_css.md
```diff
@@ -0,0 +1,153 @@
...
+}
+.module-b {
+	//@media (min-width: 768px) {
+	@include media-breakpoint-up(sm) {
+		flex-order: 1;
+	}
+}
+```
+
+Here, you can find more reasons and tips on How to write HTML(link)
```
jbalsas: What's the here link here? :)

css/how_to_write_responsive_css.md
```diff
@@ -0,0 +1,153 @@
...
+		z-index: 10; // z-index should be a variable
+	}
+}
+```
+
+In this example you can see how sometimes we recommend to use `@media (max-width: value) {` when it creates a lighter code
+
+## 3. What about the HTML?
+
+From the beginning CSS and HTML have always been the perfect _couple_, this time is not different.
```
jbalsas: Hey @wincnt, does this make sense in english? is there a different saying for it? I know you can say _a match made in heaven_ but not sure if the perfect couple makes sense.

I also think it might be "Since the beginning of time..."?

wincent: Wearing my English teacher hat, I'd say:

- We shouldn't emphasize "couple" with italics as it carries no special weight in the sentence.
- "this time is no different" would be more idiomatic than "this time is not different".
- You're right that "match made in heaven" is an idiom, but it is probably overstating things a bit 😂 — we could tone this whole thing down a bit without making claims about "From the earliest moments after the Big Bang..."

marcoscv-work: @eduardoallegrini

eduardoallegrini:
https://github.com/liferay/liferay-frontend-guidelines/pull/111

PR #91, opened 2019-09-23
docs: add link to liferay-changelog-generator

wincent:


wincent:
Merged that one because it was trivial.

PR #92, opened 2019-09-23
docs: mention liferay-js-publish

wincent:
We're using this in two places now, so let's document it.


wincent:
Merged that one because it is simple enough.

PR #102, opened 2019-10-09
docs: add general/package_structure.md

wincent:
Trying to gather together some common patterns that we use in multiple places, and have something that we can reference instead of scattered comments across GitHub like the ones I made [here](https://github.com/liferay/clay/pull/2605) and elsewhere.


jbalsas:
LGTM! Should we maybe add a comment about keeping `devDependencies` in the root of monorepos?

wincent:
> LGTM! Should we maybe add a comment about keeping `devDependencies` in the root of monorepos?

Ok, added those in https://github.com/liferay/liferay-frontend-guidelines/pull/102/commits/abdcef1c704895f05dea91724079567037bec26d

PR #105, opened 2019-10-25
docs: add debugging.md

wincent:
Just seeding this document with some initial content: stuff on logging, and in general how to make use of `NODE_ENV` to add dev-only code.


wincent:
tumbleweed

I'll just go ahead and merge this and can incorporate any feedback that comes in later.

PR #110, opened 2019-11-06
docs: Add reference to react-typescript-cheatsheet resource

bryceosterhaus:
I found this repo recently and it was pretty helpful to see other examples of TS + react. I thought it would be helpful for us to provide a link for additional resoruces.

Also, the maintainers are pretty active, so thats a plus.


wincent:
Good idea. I forced pushed to add the "docs:" prefix to the commit message to keep the Semantic PR bot happy.

PR #122, opened 2020-02-17
docs: add considerations before adding new dependencies

wincent:
File this one under "Rants".


wincent:
Will merge this and address any feedback that comes along later.

bryceosterhaus:
Glad you added this. The commands to check sizes were helpful in following along. 

As for lerna specifically, I was just looking into removing it last week. The one part I found hard to get rid of was the command `lerna version --conventional-commits`. I haven't looked too deep into this, but if we could do that without lerna, we would be fine with out lerna in clay.

jbalsas:
bla bla bla

jbalsas:
(sorry, couldn't resist it 😂)

PR #124, opened 2020-03-02
docs: react-dnd guidelines

carloslancha:


general/react_dnd.md
```diff
@@ -0,0 +1,170 @@
+# React Dnd
+
+react-dnd, wether is a good and very flexible drag and drop library for react, has one of the worst documentations ever. Anyway this is not intended to be a full replacement of reat-dnd docs but just some guides and tips to use it on a react application via hooks. For full API reference visit https://react-dnd.github.io/react-dnd/
```
wincent: > wether is a good and very flexible drag and drop library

Spelling/grammar: "whether it's", but I think you actually want something like "although a good and..."

Also, "react" should be "React".

> has one of the worst documentations ever

Grammar: "has some of the worst documentation ever"

Having said that, I don't think we should be making negative comments about a third-party library in publicly in this tone. Can we go for something a little less judgemental? Maybe:

> react-dnd is a powerful and flexible drag and drop library for React, but there are some gaps in its documentation.

Also:

> Anyway this is not intended to be a full replacement of reat-dnd docs

- "Anyway" need a comma after it.
- "reat-dnd" → "react-dnd"

> to use it on a react application via hooks

"react" → "React"

general/react_dnd.md
```diff
@@ -0,0 +1,170 @@
...
+    <DndProvider backend={Backend}>
+      /* your drag-and-drop application */
+    </DndProvider>
+  )
+}
+```
+
+## useDrag
+
+First thing you need to configure is your draggable elements using useDrag hook.
```
wincent: > using useDrag hook

Suggest:

> using the `useDrag` hook

general/react_dnd.md
```diff
@@ -0,0 +1,170 @@
...
+- `Index 1`: A connector function for the drag source. This **must** be attached to the draggable portion of the DOM.
+- `Index 2`: A connector function for the drag preview. This may be attached to the preview portion of the DOM in case you want to overwrite the default drag preview behavior.
+
+### Hook config object (most important properties)
+
+- `item`: Required. This is the only information about the draggable item that will be available for the drop or custom drag layer objects, so you need to place here any information you'll need on your drop logic.
+
+    `item.type` is also required, only drop targets registered for the same type will react to this drag.
+
+- `collect`: Optional. This is a function that receives `monitor` and `props`, and must return a plain object that you'll receive as the fist array element of the hook return value.
```
wincent: "fist" → "first"

general/react_dnd.md
```diff
@@ -0,0 +1,170 @@
...
+
+### Connect drag with DOM element
+
+You can connect drag with an element in two differente ways:
```
wincent: "differente" → "different"

general/react_dnd.md
```diff
@@ -0,0 +1,170 @@
...
+
+```
+return (
+    <div ref={drag}>
+        ...
+    </div>
+);
+```
+
+- Creating your own `ref`, calling `drag` returned parameter passing it out this ref and setting your element's ref with it.
```
wincent: This sentence doesn't sound quite right. Suggest either:

> Creating your own ref, passing it to `drag()`, and using the ref on the element.

(Note that you don't need backticks around "ref"; in [the React docs](https://reactjs.org/docs/refs-and-the-dom.html), "Ref" is just a normal noun — the variable name itself doesn't matter).

Or just:

> Creating your own ref.

(Because the code same makes it very obvious what you should do with it.)

general/react_dnd.md
```diff
@@ -0,0 +1,170 @@
...
+
+`useDrop` returns an array of two values:
+
+- `Index 0`: An object containing collected properties from the collect function. If no collect function is defined, an empty object is returned.
+
+- `Index 1`: A connector function for the drop target. This **must** be attached to the drop-target portion of the DOM.
+
+### Hook config object (most important properties)
+
+- `accept`: Required. This must be set with the item type you want drop to react to.
```
wincent: Hehe, I saw "react" here and immediately though you meant "React". Maybe "accept" would be better than "react to"?

general/react_dnd.md
```diff
@@ -0,0 +1,170 @@
...
+
+- `collect`: Optional. This is a function that receives `monitor` and `props`, and must return a plain object that you'll receive as the fist array element of the hook return value.
+
+- `drop`: Called when a compatible item is dropped on the target.
+
+- `hover`: Called when an item is hovered over the component. You can check `monitor.isOver({ shallow: true })` to test whether the hover happens over just the current target, or over a nested one. Unlike drop(), this method will be called even if `canDrop()` is defined and returns false. You can check `monitor.canDrop()` to test whether this is the case.
+
+### Connect drop with DOM element
+
+You can connect drop with an element in two differente ways:
```
wincent: "differente" → "different"

general/react_dnd.md
```diff
@@ -0,0 +1,170 @@
...
+
+```
+return (
+    <div ref={drop}>
+        ...
+    </div>
+);
+```
+
+- Creating your own `ref`, calling `drop` returned parameter passing it out this ref and setting your element's ref with it.
```
wincent: Same issue as above in the "Connect drag with DOM element" section. Whatever sentence you go with up there, you should do the same here.

wincent:
Looks good, thanks for the fixed. Feel free to merge.

Issue #133, opened 2020-03-24
Stop checkLinks.js from freaking out about codepen.io links


wincent:
As noted [here](https://github.com/liferay/liferay-frontend-guidelines/pull/132#issuecomment-603322554), our checkLinks.js can't verify codepen.io links:

```
dxp/how_to_use_css_custom_properties.md: Status code 403 for remote link: https://codepen.io/eduardoallegrini/pen/WNvgWMp
dxp/how_to_use_css_custom_properties.md: Status code 403 for remote link: https://codepen.io/eduardoallegrini/pen/gOpdyKP
dxp/how_to_use_css_custom_properties.md: Status code 403 for remote link: https://codepen.io/eduardoallegrini/pen/vYEGYZK
```

If you fetch it with `curl`, you see the cause in the returned HTML:

> Please enable cookies.
> One more step
> Please complete the security check to access codepen.io
> Why do I have to complete a CAPTCHA?
>
> Completing the CAPTCHA proves you are a human and gives you temporary access to the web property.
> What can I do to prevent this in the future?
>
> If you are on a personal connection, like at home, you can run an anti-virus scan on your device to make sure it is not infected with malware.
>
> If you are at an office or shared network, you can ask the network administrator to run a scan across the network looking for misconfigured or infected devices.
>
> Cloudflare Ray ID: 5791963dfe2fc693 • Your IP: 95.122.138.127 • Performance & security by Cloudflare

PR #134, opened 2020-03-24
fix: stop checkLinks.js from freaking out about codepen.io links

wincent:
As explained in the related issue, Cloudflare rejects our link checks with a 403 and an HTML page that asks us to fill out a CAPTCHA. I checked to see if some lighter-weight check like an OPTIONS request avoided the issue, but it doesn't.

So, grabbed some headers from the Firefox network pane and this seems to be a reasonably minimal not-robot-like set that lets the requests succeed.

Closes: https://github.com/liferay/liferay-frontend-guidelines/issues/133

### DIFF
--- a/support/checkLinks.js
+++ b/support/checkLinks.js
@@ -12,6 +12,22 @@ const statAsync = promisify(fs.stat);
 
 const FILTER_PATTERN = /\.md$/;
 
+/**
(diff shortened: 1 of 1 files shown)


wincent:
Seeing as nobody runs `yarn ci` but me, only I will be hurt if this is broken, and I don't think it is, so I am going to merge it.

PR #140, opened 2020-03-27
docs: add wysiwyg.md

mateomustapic:

PR #142, opened 2020-04-02
fix: teach checkLinks.js to handle more heading and link types

wincent:
As discussed [here](https://github.com/liferay/liferay-frontend-guidelines/pull/140#discussion_r402427606), we have some more challenging heading and link types to handle in that doc; eg.

- Headings with colons.
- Headings with backticks.
- Headings with links.
- Combinations of the above.
- Links whose case doesn't match the case of the GitHub-generated anchor.
- Links that use a novel angle-bracket syntax (eg. `[text](<http://example.com>)`).

Examples:

    Headings:

    ### [`EditorConfigContributor`](https://github.com/liferay/liferay-portal/blob/61601e89b64240db742eceaf82e86460620bcd97/portal-kernel/src/com/liferay/portal/kernel/editor/configuration/EditorConfigContributor.java#L105-L130)

    #### `public String getResourcesJspPath`

    Links:

    of the new [JavaScript APIs](#JavaScript)

    (actual anchor is #javascript)

    [text](<http://example.com>)

### DIFF
--- a/support/checkLinks.js
+++ b/support/checkLinks.js
@@ -176,6 +176,11 @@ function extractLinks(contents, file) {
 			links.add(link);
 			return ' ';
 		})
+		// [link text](<https://example.com>)
+		.replace(/\[[^\]\n]+\]\(<([^\s>]+)>\)/g, (_, link) => {
+			links.add(link);
+			return ' ';
+		})
 		// [link text](https://example.com)
 		.replace(/\[[^\]\n]+\]\(([^\s)]+)\)/g, (_, link) => {
 			links.add(link);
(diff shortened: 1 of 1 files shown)


wincent:
Force push was just to split it into two commits because we're really fixing two different things here.

support/checkLinks.js
```diff
@@ -49,11 +49,27 @@ async function checkInternal(link, files) {
...
+					return text;
+				})
+				.replace(/[ -,]+/g, '-')
+				.replace(/[`:]/g, '');
+
+			// Lowercase because the browser matches case-insensitively.
+			targets.add(target.toLowerCase());
+
+			return match;
+		});
```
wincent, line 71: I'm aware that this isn't super efficient. Choosing not to prematurely optimize here because the lack of efficiency isn't actually a problem (number of documents is so small), and that way we can keep things simple. Full check (including network access to check remote links, and `yarn` start-up time) takes 3.6s.

wincent:
Going to merge and will follow-up if necessary.

PR #144, opened 2020-04-03
docs: get rid of "dxp/architecture", for now

wincent:
This undoes the move done in 02ea9159da12f4ee6f078a1edb5bfe54d93e111a:

> docs: moves wysiwyg.md to a specific folder to differentiate architecture explanations from coding best practices

At the moment, there is only one "architecture" document, and I think a good rule of thumb is that we don't put anything into a folder unless there are (at least) three things that go together in it.

Note that I'm not saying we should never have an "architecture" folder, just that we shouldn't have one yet.

Before this commit:

```
css
├── how_to_name_the_css_classes.md
├── how_to_name_the_css_files.md
├── how_to_write_responsive_css.md
├── when_and_how_to_create_a_new_css_component.md
└── when_to_use_css_utility_classes.md
dxp
├── architecture
│   └── rich_text_editors.md
├── aspect_oriented_programming.md
├── browser_detection.md
├── debounce_and_throttle.md
├── debugging.md
├── dev_dependencies.md
├── drag_and_drop.md
├── environment.md
├── formatting.md
├── how_to_use_css_custom_properties.md
├── linting.md
├── requests.md
└── urls.md
general
├── branch_structure.md
├── comments.md
├── commit_messages.md
├── file_names.md
├── formatting.md
├── imports.md
├── javascript.md
├── package_structure.md
├── pull_requests.md
├── react.md
├── react_dnd.md
├── security.md
├── testing
│   ├── README.md
│   ├── anatomy.md
│   ├── supported_libraries.md
│   └── ui_components.md
└── typescript.md
```

We should avoid the temptation to categorize and hierarchize everything, because until the number of items in a folder exceeds a threshold (of say, 20 items), deepening the hierarchy may not actually help; it may just makes things harder to find, and harder to know where to place.

It's not clear to me why "rich_text_editors" is any different than, say, "drag_and_drop". Both of these, as well as others, basically fit into the category of "how to do stuff" in DXP. In the past, we had folders named "guidelines" and "patterns" (we de-hierarchized those in 520d67b92681d). All of these schemes suffer from the problem that the lines between categories are blurry and coming up with non-overlapping, unambiguous, redundancy-free taxonomies is hard. It's often best to go with something "dumb" and "shallow" until the size obliges you to take another approach.


wincent:
Ok, Chema is out sick, so this goes in, like a thief in the night!

thief

bryceosterhaus:
two great things about this gif.

1.kicking the door like a ninja on the way out
2. the way she holds the gun.

Top notch gif

jbalsas:
treason

Issue #80, opened 2019-09-02
Create a document with all the mixins, functions and variables available in the Portal and Clay and how to use them


eduardoallegrini:
One of the problem I encounter during my latest tasks was the lacking (or apparent absence) of a document where to find mixins, variables and functions altogether.

It could be useful to all the developers, to have a Document (in list or table format) where we can gather these values with a short description or instruction about its use.

Of course we can search through the Clay and Portal files but sometimes there are deprecated results that be confusing (compass, bourbon, bs3_compat_layer, etc)

We should also add the simple information about the imported-clay-files
example:
`@import atlas-variables.scss;`
"what does it means?"
"how can I use it?"
"why I should use it instead of `base-variables` or `clay-atlas`?"

@marcoscv-work, @pat270, @jbalsas What do you think?

eduardoallegrini:
Eventually we should include this https://sass-lang.com/documentation/functions

matuzalemsteles:
I'm afraid this is a bit costly to maintain, @pat270 is using [sassdoc](http://sassdoc.com/) to generate these documents in Clay, we are thinking of adding this to the site.

pat270:
I agree with @matuzalemsteles. I feel like it will get out of date quickly. Regardless, the documentation should be focused on DXP related parts instead of Sass + Compass + Bourbon + Bootstrap + Clay CSS + DXP.

jbalsas:
I think sassdocs looks quite nice and useful @pat270! If we could also bring in those coming from bootstrap it'd be great.

There's one thing that we might want to think about. In my opinion, autogenerated docs are great for reference, and because they're easy to maintain and keep updated. They, however, are not curated at all and lack a clear intent and messaging.

That is, providing reference docs is like doing a brain-dump over our developers. Many of the mixins and variables are mostly for internal consumption. I feel we should make a curated list of what mixins we consider useful and common practice so we can promote their proper usage and explain how they're good to solve specific issues.

pat270:
Thanks @jbalsas! I can document the Bootstrap stuff, just need to restructure the docs a bit. I agree with having a curated list, but I question whether it should all live in this repo. I think best practices related to Clay should live in the Clay repo, that way it's less likely to get out of sync.

jbalsas:
Yeah, agreed, those should live in Clay. From here we can simply link to there

PR #148, opened 2020-04-06
docs: Adds Clay CSS 2 Recipes Getting Started

pat270:
issue #80

Instead of trying to figure out all the questions devs may have about Clay CSS, I thought documenting the framework through a cookbook format might be better answering questions that may arise. I wasn't sure where to put it, let me know if there's a better place and anything that needs more clarification.

### DIFF
--- a/css/clay_css_2_recipes/02_buttons.md
+++ b/css/clay_css_2_recipes/02_buttons.md
@@ -1,0 +1,445 @@
+# Buttons
+
+Clay CSS follows Bootstrap 4's base + modifier pattern, similar to [OOCSS](https://github.com/stubbornella/oocss/wiki). The base class, `.btn`, should only contain styles common to all `.btn` elements, such as `border-style`, `border-width`, `display`, `font-size`, and any other `button` related css resets.
+
+The modifier, `.btn-primary`, provides the styles for the main variant of `.btn`, such as `background-color`, `border-color`, and `color`. The modifier, `.btn-secondary`, provides styles for the next `.btn` variant.
+
+This recipe will modify the button classes in Clay CSS to match Material Design's shown in [https://material.io/components/buttons/](https://material.io/components/buttons/). The ripple effect on click requires external javascript and is outside the scope of this framework.
+
+## Base Button
+
+Material Design's base button styles:
+
+```
+.mdc-button {
+	-webkit-appearance: none;
+	align-items: center;
+	border-radius: 4px;
+	border: none;
+	box-sizing: border-box;
+	display: inline-flex;
+	font-family: Roboto, sans-serif;
+	font-size: .875rem;
+	-moz-osx-font-smoothing: grayscale;
+	-webkit-font-smoothing: antialiased;
+	font-weight: 500;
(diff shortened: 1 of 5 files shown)


css/clay_css_2_recipes/02_buttons.md
```diff
@@ -0,0 +1,445 @@
+# Buttons
+
+Clay CSS follows Bootstrap 4's base + modifier pattern, similar to [OOCSS](https://github.com/stubbornella/oocss/wiki). The base class, `.btn`, should only contain styles common to all `.btn` elements, such as `border-style`, `border-width`, `display`, `font-size`, and any other `button` related css resets.
+
+The modifier, `.btn-primary`, provides the styles for the main variant of `.btn`, such as `background-color`, `border-color`, and `color`. The modifier, `.btn-secondary`, provides styles for the next `.btn` variant.
+
+This recipe will modify the button classes in Clay CSS to match Material Design's shown in [https://material.io/components/buttons/](https://material.io/components/buttons/). The ripple effect on click requires external javascript and is outside the scope of this framework.
```
wincent, line 7: nit: javascript → JavaScript

css/clay_css_2_recipes/02_buttons.md
```diff
@@ -0,0 +1,445 @@
...
+
+.btn-unstyled {
+	letter-spacing: normal;
+	text-transform: none;
+}
+```
+
+## Button Variants
+
+Material Design uses a white overlay and opacity with javascript to create the ripple effect on click. We will try to match the `:active` color as close as possible.
```
wincent, line 104: "JavaScript" here too.

matuzalemsteles:
hey @pat270 it would be great to add that to clayui.com 😁

wincent:
Looks like this needs a `yarn format`. The link checker is also failing to verify those font links (`https://fonts.googleapis.com/css2?family=Roboto:wght@300;400;500;700&display=swap`) even though they are valid. It is returning a 400 status, probably sniffing something "bot-like" about the request headers. If we can't tweak the headers, we could just add a blacklist to the link checker to avoid checking this kind of link — the intention isn't that literally every URL in the document should be correct; it's that links that a user might actually click on should lead somewhere (not so much the case for these font links, of which only one of them is even clickable and perhaps it shouldn't even be). I'll look into it and report back.

About where this should go: I'm open to it going here or anywhere. Even existing is already valuable.

wincent:
@pat270: I pushed some commits on top to fix CI (formatting and bad link check). I'm going to merge this because CI is green, so that it doesn't sit in limbo while we decide whether this or clayui.com is the right home for it. (I suspect that getting it into clayui.com will take a bit longer. But if we do move it there, we can link to it there from here easily enough.)

pat270:
Thanks @wincent @matuzalemsteles, we can drop them in the Clay docs site.  It doesn't matter to me where, maybe we can hash out the docs here first? BTW this is for 2.x.

matuzalemsteles:
@pat270 no problem, we can do that.

PR #163, opened 2020-05-29
docs: add links to Liferay-specific editor integration info

wincent:
As pointed out [here](https://github.com/markocikos/liferay-portal/pull/188#issuecomment-635818859), the info here was stale and did not reflect [the latest editor integration guidelines](https://github.com/liferay/liferay-npm-tools/tree/master/packages/liferay-npm-scripts#editor-integrations).

Rather than copy them here (and risk them getting out of sync), I just linked to them.

Note: CI is going to fail on this because CloudFlare has decided to 503 the link verification requests (again). I'll fix that in a separate pull.

PR #161, opened 2020-05-20
docs: Clay CSS 2 Recipes Getting Started Source formatting; Word smith; Update grammar

pat270:

PR #164, opened 2020-05-29
fix: give up on trying to check codepen.io links

wincent:
Since https://github.com/liferay/liferay-frontend-guidelines/pull/161 Cloudflare has been 503-ing these links.

They used to 403 them, and we got around that in https://github.com/liferay/liferay-frontend-guidelines/pull/134, but it looks like the party is over now, and I don't think it is worth trying to keep up with them on this one. [Another example of a PR](https://github.com/liferay/liferay-frontend-guidelines/pull/163) with a
spurious failure due to these links.

### DIFF
--- a/support/checkLinks.js
+++ b/support/checkLinks.js
@@ -13,8 +13,8 @@ const statAsync = promisify(fs.stat);
 const FILTER_PATTERN = /\.md$/;
 
 /**
- * Need not-robot-like headers to prevent Cloudflare from rejecting our
- * external link checks with 403s.
+ * Use not-robot-like headers to attempt to prevent Cloudflare from
+ * rejecting our external link checks with 403 or 503 status.
  *
  * See: https://github.com/liferay/liferay-frontend-guidelines/issues/133
  */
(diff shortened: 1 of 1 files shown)


support/checkLinks.js
```diff
@@ -13,8 +13,8 @@ const statAsync = promisify(fs.stat);
 const FILTER_PATTERN = /\.md$/;
 
 /**
- * Need not-robot-like headers to prevent Cloudflare from rejecting our
- * external link checks with 403s.
+ * Use not-robot-like headers to attempt to prevent Cloudflare from
+ * rejecting our external link checks with 403 or 503 status.
```
wincent, line 17: Just to clarify why I'm not deleting this outright. There is still a chance that these headers *might* allow some Cloudflare-hosted links to be checked, because it seems the security levels are configurable on a per-site basis. For example, see the responses here:

- https://stackoverflow.com/a/18523671/2103996
- https://community.cloudflare.com/t/trying-to-curl-to-a-site-protected-by-cloudflare/32918

In short, depends on whether IPs are whitelisted, protections are turned on or off etc.

wincent:
Too trivial to warrant review.

PR #168, opened 2020-07-02
docs(code review): add gotcha note about notifications

wincent:
I continue to see people creating PRs without descriptions, and then editing them to create (great!) descriptions, but people don't get the notification.


izaera:
I believe it happens in edited comments too. You only get the first edit.

Nice addition. We should probably notify it in slack too? :thinking:

jbalsas:
> This is true for `@`-mentions added when editing comments

👆 

LGTM! We already went overt his in our team slack, I think? it won't hurt going over it again... I know I always forget about it :(

wincent:
> LGTM! We already went overt his in our team slack, I think? it won't hurt going over it again... I know I always forget about it :(

Yep, the reason I added it here is that I saw it happening again yesterday. I think it will take some repetition for people to realize/remember it.

I'll fix formatting and merge.

PR #169, opened 2020-07-07
docs: add some additional notes on file naming (constants)

wincent:
Seeing as this came up [in review here](https://github.com/liferay-frontend/liferay-portal/pull/48#discussion_r450509707) I thought we may as well document it. Specifically, the question is how to name a file that exports a bunch of constants.

We have a tension between what our simple rule ("file names should match their contents") would imply, and what we are actually doing in liferay-portal.

So, given that this document is somewhat aspirational in nature (ie. describing what we should do even if it doesn't yet match what we actually do), I've noted that we should be doing (use all caps in the name) and also what we tend to do in liferay-portal up to today... Maybe one day we will be able to change this.

Note that there are still two approaches to defining constants that we could debate, but given that we currently seem to be doing it the same way (the first one), I didn't get into that in this PR:

1.  Export default object:

    ```js
    // in COLORS.js:
    const BLUE = '#00f';
    const GREEN = '#0f0';
    const RED = '#f00';

    export default COLORS = {BLUE, GREEN, RED};

    // or:
    const COLORS = {BLUE, GREEN, RED};
    export default COLORS;

    // or even:
    export default {BLUE, GREEN, RED};

    // in other.js:
    import COLORS from './COLORS';
    ```

2.  Export individual items:

    ```js
    // in COLORS.js:
    export const BLUE = '#00f';
    export const GREEN = '#0f0';
    export const RED = '#f00';

    // in other.js:
    import * as COLORS from './COLORS';
    ```

### DIFF
--- a/general/file_names.md
+++ b/general/file_names.md
@@ -4,15 +4,16 @@
 
 ## Examples
 
-| File contents                                 | File name                   |
-| --------------------------------------------- | --------------------------- |
-| A class, `ThingParser`                        | `ThingParser.js`            |
-| A component, `GizmoPicker`                    | `GizmoPicker.js`            |
-| A constructor function, `Blob`                | `Blob.js`                   |
-| A function, `concatenateStrings`              | `concatenateStrings.js`     |
-| Multiple exports, one of which is the default | Based on the default export |
-| Multiple exports, with no default             | Descriptive CamelCase       |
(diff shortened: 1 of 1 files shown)


wincent:
Ok, so this is good enough for me:

Screenshot 2020-07-08 -135145-PCHzilLh@2x

I'm not even one of the two!

PR #191, opened 2020-09-04
docs: document frontend-taglib-clay and it's relation with @clayui components

bryceosterhaus:
I wasn't sure how in depth we want to go with these. I initially had more content but it felt over-wordy. Let me know if there are any other caveats or nuances to note with this.

### DIFF
--- a/dxp/frontend_taglib_clay.md
+++ b/dxp/frontend_taglib_clay.md
@@ -1,0 +1,49 @@
+# What is `frontend-taglib-clay` and it's relation with clay-\*, @clayui, and css?
+
+## High-level:
+
+`frontend-taglib-clay` provides a collection of clay dependencies from npm and custom components to be used across DXP as taglibs and js dependencies. Versions of these npm modules are locked in this module, this is done to keep modules consistent across dxp. For example, by locking these versions, this means we use the same version of `@clayui/button` across all of dxp so that we can reduce the number of modules served with dxp.
+
+## How is it structured:
+
+`resources/META-INF/resources`: These are the custom JS components
+`resources/META-INF/liferay-clay.tld`: This is the list of taglib definitions that can be used across DXP.
+`package.json`: Lists all the provided versions of `clay-*` and `@clayui/*` packages.
+
+## How does it work:
+
+For example, if you want to use the version of `@clayui/button` specified in `liferay-portal` or a clay taglib, you need to declare a gradle dependency on `frontend-taglib-clay`. For example
+
+build.gradle
+
+```gradle
+dependencies {
+	//....
+	compileOnly project(":apps:frontend-taglib:frontend-taglib-clay")
+}
+```
+
+After declaring the gradle dependency, you then can use one of the clay npm modules like so...
+
+-   `package.json` - Add the dependency you want, for example `"@clayui/button": "3.4.0"`
+-   Use the dependency as normal within your JS components
+
+For using the taglib
+
+-   `init.jsp` - Add tag dependency `<%@taglib uri="http://liferay.com/tld/clay" prefix="clay" %>`
+-   Use as normal in jsp
+
+```java
+<clay:button
+	displayType="secondary"
+	icon="view"
+/>
+```
+
+### Caveats
+
+Some of the taglib definitions provide the `propsTransformer` property. This prop is unique to dxp and gives the user the ability to manipulate data that is provided to the react component within the taglib. See [content-dashboard](https://github.com/liferay/liferay-portal/blob/master/modules/apps/content-dashboard/content-dashboard-web/src/main/resources/META-INF/resources/view.jsp#L207) for example of use.
+
+## Updating `frontend-taglib-clay`:
+
+Whenever there is a new version of `@clayui/*` or `clay-*` modules, we update all the versions in this module so that the most up to date version is used across dxp.
--- a/dxp/frontend_taglib_clay.md
+++ b/dxp/frontend_taglib_clay.md
@@ -1,20 +1,20 @@
-# What is `frontend-taglib-clay` and it's relation with clay-\*, @clayui, and css?
+# What is `frontend-taglib-clay` and its relation with clay-_, @clayui/_, and clay-css?
 
 ## High-level:
 
-`frontend-taglib-clay` provides a collection of clay dependencies from npm and custom components to be used across DXP as taglibs and js dependencies. Versions of these npm modules are locked in this module, this is done to keep modules consistent across dxp. For example, by locking these versions, this means we use the same version of `@clayui/button` across all of dxp so that we can reduce the number of modules served with dxp.
+`frontend-taglib-clay` provides a collection of Clay dependencies from npm and custom components to be used across DXP as taglibs and JS dependencies. Versions of these npm modules are locked in this module, this is done to keep modules consistent across DXP. By locking these versions, this means we use the same version of `@clayui/button` across all of DXP so that we can reduce the number of modules served with DXP.
 
-## How is it structured:
+## Structure:
 
-`resources/META-INF/resources`: These are the custom JS components
-`resources/META-INF/liferay-clay.tld`: This is the list of taglib definitions that can be used across DXP.
-`package.json`: Lists all the provided versions of `clay-*` and `@clayui/*` packages.
+-   `resources/META-INF/resources`: These are the custom JS components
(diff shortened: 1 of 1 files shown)


dxp/frontend_taglib_clay.md
```diff
@@ -0,0 +1,49 @@
+# What is `frontend-taglib-clay` and it's relation with clay-\*, @clayui, and css?
```
wincent, line 1: it's → its

dxp/frontend_taglib_clay.md
```diff
@@ -0,0 +1,49 @@
...
+
+## High-level:
+
+`frontend-taglib-clay` provides a collection of clay dependencies from npm and custom components to be used across DXP as taglibs and js dependencies. Versions of these npm modules are locked in this module, this is done to keep modules consistent across dxp. For example, by locking these versions, this means we use the same version of `@clayui/button` across all of dxp so that we can reduce the number of modules served with dxp.
```
wincent, line 5: Let's always capitalize:

- "clay" → Clay
- "dxp" → DXP
- "npm" → NPM
- "js" → JS

I know you were worried about word count, so you could delete "For example," and just start that sentence with "By locking".

dxp/frontend_taglib_clay.md
```diff
@@ -0,0 +1,49 @@
...
+
+## How is it structured:
```
wincent, line 7: If this is a question, end with "?" and not ":"

(Otherwise, reword as a non-question;  ie. "How it is structured".)

dxp/frontend_taglib_clay.md
```diff
@@ -0,0 +1,49 @@
...
+
+`resources/META-INF/resources`: These are the custom JS components
+`resources/META-INF/liferay-clay.tld`: This is the list of taglib definitions that can be used across DXP.
+`package.json`: Lists all the provided versions of `clay-*` and `@clayui/*` packages.
```
wincent, line 11: These don't format correctly. Make them bullet points?

dxp/frontend_taglib_clay.md
```diff
@@ -0,0 +1,49 @@
...
+
+## How does it work:
```
wincent, line 13: If it's a question, change ":" → "?" — or just turn it into a non-question ("How it works").

dxp/frontend_taglib_clay.md
```diff
@@ -0,0 +1,49 @@
...
+
+For example, if you want to use the version of `@clayui/button` specified in `liferay-portal` or a clay taglib, you need to declare a gradle dependency on `frontend-taglib-clay`. For example
```
wincent, line 15: Capitalize:

- Clay
- Gradle

I think you could avoid the repeated "For example" by just omitting the second one.

dxp/frontend_taglib_clay.md
```diff
@@ -0,0 +1,49 @@
...
+
+build.gradle
```
wincent, line 17: Maybe this should be a heading:

```md
### `build.gradle`
```

dxp/frontend_taglib_clay.md
```diff
@@ -0,0 +1,49 @@
...
+
+```gradle
+dependencies {
+	//....
+	compileOnly project(":apps:frontend-taglib:frontend-taglib-clay")
+}
+```
+
+After declaring the gradle dependency, you then can use one of the clay npm modules like so...
```
wincent, line 26: Capitalize:

- Gradle
- Clay
- NPM

Probably want ":" instead of "..."

dxp/frontend_taglib_clay.md
```diff
@@ -0,0 +1,49 @@
...
+
+-   `package.json` - Add the dependency you want, for example `"@clayui/button": "3.4.0"`
```
wincent, line 28: Maybe instead of "for example" → "in this case,"

dxp/frontend_taglib_clay.md
```diff
@@ -0,0 +1,49 @@
...
+-   Use the dependency as normal within your JS components
+
+For using the taglib
```
wincent, line 31: End this one with a colon?

dxp/frontend_taglib_clay.md
```diff
@@ -0,0 +1,49 @@
...
+
+-   `init.jsp` - Add tag dependency `<%@taglib uri="http://liferay.com/tld/clay" prefix="clay" %>`
+-   Use as normal in jsp
```
wincent, line 34: Capitalize: "jsp" → JSP

dxp/frontend_taglib_clay.md
```diff
@@ -0,0 +1,49 @@
...
+```java
+<clay:button
+	displayType="secondary"
+	icon="view"
+/>
+```
+
+### Caveats
+
+Some of the taglib definitions provide the `propsTransformer` property. This prop is unique to dxp and gives the user the ability to manipulate data that is provided to the react component within the taglib. See [content-dashboard](https://github.com/liferay/liferay-portal/blob/master/modules/apps/content-dashboard/content-dashboard-web/src/main/resources/META-INF/resources/view.jsp#L207) for example of use.
```
wincent, line 45: Capitalize:

- DXP
- React

I know that @jbalsas said he was going to take a stab at explaining the `propsTransformer` stuff in more depth, so I am not sure whether that would be in a separate document or contributing to this one. 🤷‍♂️

dxp/frontend_taglib_clay.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Updating `frontend-taglib-clay`:
+
+Whenever there is a new version of `@clayui/*` or `clay-*` modules, we update all the versions in this module so that the most up to date version is used across dxp.
```
wincent, line 49: Capitalize "DXP".

I also think it would be worth documenting _how_ we do the update (if I recall correct, there is some Gradle task that will force the version numbers into sync across all OSGi modules but it's been so long since I had to use it that I can't remember whether it's just a simple `formatSource` from the top).

wincent (commented):
> I wasn't sure how in depth we want to go with these. I initially had more content but it felt over-wordy. Let me know if there are any other caveats or nuances to note with this.

It's hard to know when to stop (I usually go too far), but it's clear that having even minimal/brief documentation is a great improvement over having no documentation at all.

I left some little copy-editing nits inline.

dxp/frontend_taglib_clay.md
julien: @wincent  not something important, but I'm never 100% sure whether to use "NPM" or "npm", for example in their [about](https://www.npmjs.com/about) page they refer to it as npm in lowercase and if you do a quick Google search for npm you'll see this (which has npm, Npm and NPM) :shrug: 

npm01

dxp/frontend_taglib_clay.md
julien: @wincent I think this is a good place to explain/document the `propsTransformer` because as far I've seen it's going to be used mostly in this taglib

dxp/frontend_taglib_clay.md
wincent: 🤦‍♂️ — yeah, seems they probably use "npm" more than they use the variants, and it looks like [their first commit in 2009](https://github.com/npm/cli/commit/4626dfa73b7847e9c42c1f799935f8242794d020) used "npm" already then.

I have historically used "NPM" for the company/registry/package format, and `npm` for the commandline tool, to distinguish them, but given what you posted I could be persuaded to do "npm" for both (sadly, means that there are quite a few of our repos with docs/references to "NPM" in them).

julien: > I have historically used "NPM" for the company/registry/package format, and npm for the commandline tool,

Makes sense to me.

izaera: I think I use `npm` too. I usually check the top left header of each project I mention to see the capitalization, so:

- npm
- webpack

However, the rule doesn't work for Babel, which they call Babel, but capitalize as BABEL in the image :shrug: .

In any case, for npm the correct form seems to be npm. Even `npm, Inc.` is capitalized like that, and look at [their blog](https://blog.npmjs.org/)

bryceosterhaus:
@wincent just updated and added your recommendations. I also added some content at the end about how the dependencies are actually updated and the steps we follow.

wincent:
> I also added some content at the end about how the dependencies are actually updated and the steps we follow.

Looks good. I'm going to amend the commit message to keep the bot happy and then merge.

PR #197, opened 2020-09-08
docs: React rendering infrastructure in DXP

izaera:


dxp/react_rendering.md
```diff
@@ -0,0 +1,148 @@
+# React Rendering in DXP
+
+This article describes DXP infrastructure to render React components from the server side perspective. This includes taglibs as well as low level services.
```
wincent, line 3: "low level" → "low-level"

dxp/react_rendering.md
```diff
@@ -0,0 +1,148 @@
...
+
+## React Renderer Service
+
+-   OSGi bundle: `com.liferay.portal.template.react.renderer.api`
+-   Interface name: `com.liferay.portal.template.react.renderer.ReactRenderer`
+-   Service type: `singleton`
+
+This is the low level service to render React components.
```
wincent, line 11: "low level" → "low-level"

dxp/react_rendering.md
```diff
@@ -0,0 +1,148 @@
...
+```java
+public void renderReact(
+		ComponentDescriptor componentDescriptor, Map<String, Object> data,
+		HttpServletRequest httpServletRequest, Writer writer)
+	throws IOException;
+```
+
+which receives:
+
+1. A `ComponentDescriptor` describing the source of the React component to render. More on this on the next section.
```
wincent, line 26: "on the next section" → "in the next section"

dxp/react_rendering.md
```diff
@@ -0,0 +1,148 @@
...
+Map<String, Object> data = new HashMap<>();
+
+data.put("color", "red");
+
+reactRenderer.renderReact(
+	componentDescriptor, data, httpServletRequest,
+	httpServletResponse.getWriter());
+```
+
+This will cause an HTML like this to be rendered in the current position of the `HttpServletResponse`:
```
wincent, line 64: "an HTML like this" → "HTML like this"

dxp/react_rendering.md
```diff
@@ -0,0 +1,148 @@
...
+
+```html
+<div id="XJAZ"></div>
+```
+
+(where `XJAZ` is a random placeholder id).
+
+And a JavaScript like this to be appended at the end of the page:
```
wincent, line 72: "a JavaScript" → "JavaScript"

dxp/react_rendering.md
```diff
@@ -0,0 +1,148 @@
...
+/>
+```
+
+## liferay-clay Tag Library
+
+-   OSGi bundle: `com.liferay.frontend.taglib.clay`
+-   Taglib URI: `http://liferay.com/tld/clay`
+-   Tags: `<liferay-clay:...>`
+
+The `liferay-clay` taglib provides several tags to render Clay components (for example: `<liferay-clay:alert>`, `<liferay-clay:button>`, ...). This is tangentially related to React because Clay components are implemented as React components.
```
wincent, line 124: Maybe not "tangentially"... more like "indirectly".

dxp/react_rendering.md
```diff
@@ -0,0 +1,148 @@
...
+```jsp
+<clay:label
+	displayType="warning"
+	label="pending"
+/>
+```
+
+```jsp
+<clay:alert
+	message="you-do-not-belong-to-an-organization-and-are-not-allowed-to-view-other-organizations"
```
wincent, line 146: A nice long one there.

wincent (commented):
lg2m, but I only pretend to understand "the Java".

dxp/react_rendering.md
izaera: What is the rule? When do you use `-` or nothing? Is there a rule, in fact?

izaera:
"the Jabba"? :thinking: 

jabba

dxp/react_rendering.md
wincent: I will probably do a bad job of explaining this, but when you have a `modifier` + `noun` which act together as though they were a single "atomic" `adjective`, you join them with a hyphen.

Looks like the official term for these is "compound adjectives":

Probably a better explanation: https://www.gingersoftware.com/content/grammar-rules/adjectives/compound-adjectives/

izaera: :-o Thanks :+1:

diegonvs:
Hey, Does make sense document about routing on this document?

I remember that routing in the (beginning of React times/old times) is pain. At Recife, We're [using react-router-dom ](https://github.com/liferay/liferay-portal/search?q=%22react-router-dom%22&unscoped_q=%22react-router-dom%22) and I remember segments [had a different way to route](https://github.com/liferay/liferay-portal/blob/3c0a0c769f5ce092b2800e859b52f5b098dd0133/modules/apps/segments/segments-web/src/main/resources/META-INF/friendly-url-routes/routes.xml)(based on the JSP paths).

jbalsas:
> Hey, Does make sense document about routing on this document?

We're not in the business of writing down good practices right now (in many cases because we don't have them yet). Our main goal right now is to simply gain knowledge of what exists in the platform so then we can try to improve things. Without that, we blindly push new things and fail to cover existing use cases and end up just bloating the platform more.

izaera:
Shall we merge this already?

jbalsas:
Let me take a quick look!

dxp/react_rendering.md
```diff
@@ -0,0 +1,148 @@
...
+
+## React Renderer Service
+
+-   OSGi bundle: `com.liferay.portal.template.react.renderer.api`
+-   Interface name: `com.liferay.portal.template.react.renderer.ReactRenderer`
+-   Service type: `singleton`
+
+This is the low-level service to render React components.
+
+> Note that it renders React components, not any React code, so you must provide a source which exports a React component to render.
```
jbalsas, line 13: Can we make this scream more? 

Or additionally, show an example of what `a source which exports a React component` means?

dxp/react_rendering.md
```diff
@@ -0,0 +1,148 @@
...
+		HttpServletRequest httpServletRequest, Writer writer)
+	throws IOException;
+```
+
+which receives:
+
+1. A `ComponentDescriptor` describing the source of the React component to render. More on this in the next section.
+2. A `Map` containing arbitrary data that will be injected to the React component as initial properties. Note that the data inside the map must be compatible with JavaScript, so you cannot inject any Java object you like, only primitive types, `Map`s, `Array`s, ... that can be serialized to JSON.
+3. The current `HttpServletRequest` for which the component is being rendered.
+4. A `Writer` where the rendered HTML (and optionally the JavaScript) is written to.
```
jbalsas, line 29: Is JS really "optional` here? We always render JS to render the React component, don't we?

dxp/react_rendering.md
```diff
@@ -0,0 +1,148 @@
...
+
+Note that parameter 4 is not intended to be used to write the rendered HTML and JavaScript to any place you like but to pass the `PrintWriter` from the `HttpServletResponse` or the `JspWriter` from the `PageContext` you have available when invoking the method.
+
+Also, take into account that, if you don't render the JavaScript inline (see option `positionInline` of `ComponentDescriptor`), the JavaScript will be written to the bottom of the current `HttpServletResponse`, while the HTML will be inlined into the passed `Writer`.
```
jbalsas, line 33: It would be interesting to link to what `HTML` is actually being rendered and why. That is, the `div` placeholder that we use to mount the component.

I get the impression from reading here that this can render arbitrary HTML (which might in the future if we server-render the components, but it's not the case now...)

dxp/react_rendering.md
```diff
@@ -0,0 +1,148 @@
...
+
+### What the ComponentDescriptor contains
+
+`ComponentDescriptor` instances are [value objects](https://en.wikipedia.org/wiki/Value_object) that describe what needs to be rendered. A `ComponentDescriptor` may have up to five fields (of which only the first one is mandatory):
```
jbalsas, line 37: Could we add links to the interfaces in the `liferay/liferay-portal` repo?

dxp/react_rendering.md
```diff
@@ -0,0 +1,148 @@
...
+
+1. `module`: the name of the AMD module that exports the React component to be rendered. Note that you need to provide the canonical name of the module (including its version number). To avoid hardcoding any version number you may use the instance of the `com.liferay.frontend.js.loader.modules.extender.npm.NPMResolver` service tied to the OSGi bundle containing the AMD module to be rendered.
```
jbalsas, line 39: Maybe add an example of this? Both in what a `module` usually looks like: `foo@1.3.2/js/Foo.js` and how to get it `NPMResolver.resolveModule('Foo')`.

dxp/react_rendering.md
```diff
@@ -0,0 +1,148 @@
...
+2. `componentId`: the `componentId` that will be given to the component's JavaScript as an initial property. Note that this field is only used if the `data` map passed to `renderReact` does **NOT** contain a `componentId` property.
+3. `dependencies`: this is deprecated; don't use it 🧐.
```
jbalsas, line 41: Rather than deprecated, I'd say this has no effect for the common usage of `ReactRenderer` since there's no ergonomic way to pass in dependencies to the React component.

I think it will be useful in some contexts where we register dynamic views (dataset-display) in a way that are available dynamically to another component via some level of indirection.

dxp/react_rendering.md
```diff
@@ -0,0 +1,148 @@
...
+4. `positionInline`: a flag to render the JavaScript directly in the current position of the HTML response or defer it to the bottom of the page. By default `false`.
+5. `propsTransformer`: optional name of an AMD module exporting a function to transform the properties before they are passed to the component.
```
jbalsas, line 43: This has to follow the same rules as the `module` attribute... just saying... 😉

dxp/react_rendering.md
```diff
@@ -0,0 +1,148 @@
...
+
+### How the React component is rendered
```
jbalsas, line 45: Love this section ❤️

dxp/react_rendering.md
```diff
@@ -0,0 +1,148 @@
+# React Rendering in DXP
+
+This article describes DXP infrastructure to render React components from the server side perspective. This includes taglibs as well as low-level services.
+
+## React Renderer Service
+
+-   OSGi bundle: `com.liferay.portal.template.react.renderer.api`
```
jbalsas, line 7: Can we link this to source code somehow?

dxp/react_rendering.md
```diff
@@ -0,0 +1,148 @@
...
+-   Interface name: `com.liferay.portal.template.react.renderer.ReactRenderer`
+-   Service type: `singleton`
```
jbalsas, line 9: Could we provide links to source code for some of these things?
- OSGi bundle
- ReactRenderer Interface
- ...

dxp/react_rendering.md
```diff
@@ -0,0 +1,148 @@
...
+);
+```
+
+If you are curious about what `portal-template-react-renderer-impl@1.2.3/render.es` contains, you can see its content in [liferay-portal's source code](https://github.com/liferay/liferay-portal/blob/master/modules/apps/portal-template/portal-template-react-renderer-impl/src/main/resources/META-INF/resources/render.es.js). It is simply a helper to connect the generated JavaScript and the React component but, in general, it's a level of indirection to be able to add any extra infrastructure code we may need in the future.
+
+## react:component Tag
+
+-   OSGi bundle: `com.liferay.frontend.taglib.react`
+-   Taglib URI: `http://liferay.com/tld/react`
+-   Tag: `<react:component>`
```
jbalsas, line 91: Same here... links to the bundle, the tld, and the component implementation?

dxp/react_rendering.md
```diff
@@ -0,0 +1,148 @@
...
+
+This tag is a wrapper for the low-level React renderer service. It leverages OSGi service `com.liferay.portal.template.react.renderer.ReactRenderer` to render the HTML and JavaScript.
+
+The tag has five attributes (all optional but the `module` one):
+
+1. `componentId`: the component identifier (which can be used to reference the React component with `Liferay.component(componentId)`).
```
jbalsas, line 97: Small nit. Remove the `(...)` and simplify the sentence maybe like: `the component identifier. It can be used to...`

Reducing subordination could make this easier to read.

dxp/react_rendering.md
```diff
@@ -0,0 +1,148 @@
...
+2. `data`: deprecated, do not use; nothing to see here 👮.
```
jbalsas, line 98: Maybe say "use `props` instead"? 😄

dxp/react_rendering.md
izaera: I will rephrase that. 

I mean that the JS gets written optionally to that Writer in the sense that it is only written if you set `positionInline` to true. Otherwise it is attached to a ScriptData and gets written to the request's Writer at the end of the request's lifecycle.

dxp/react_rendering.md
```diff
@@ -0,0 +1,148 @@
...
+	%>"
+	servletContext="<%= application %>"
+/>
+```
+
+## liferay-clay Tag Library
+
+-   OSGi bundle: `com.liferay.frontend.taglib.clay`
+-   Taglib URI: `http://liferay.com/tld/clay`
+-   Tags: `<liferay-clay:...>`
```
jbalsas, line 122: Link this for future us!

jbalsas:
Screen Shot 2020-09-10 at 11 51 53

👀

jbalsas:
Hey @izaera, this looks great to me! I left some comments here and there, feel free to address the ones you think are relevant and merge this! LG(reat)TM overall!

dxp/react_rendering.md
```diff
@@ -0,0 +1,148 @@
...
+	httpServletResponse.getWriter());
+```
+
+This will cause HTML like this to be rendered in the current position of the `HttpServletResponse`:
+
+```html
+<div id="XJAZ"></div>
+```
+
+(where `XJAZ` is a random placeholder id).
```
jbalsas, line 70: I'm missing a bit of an explanation about this placeholder element. We use it to mount the react component in the parent element, which is done through the `render` method. This is a tricky aspect of it that we should collect here, I think.

dxp/react_rendering.md
izaera: It's described some lines below.

dxp/react_rendering.md
izaera: :thinking: I said deprecated because it is not used anywhere in the code. However, reading your comment it looks more like a  _reserved parameter_, so I'll change it.

izaera: Changed to: `3. `dependencies`: this is a reserved parameter which can be used in the future to declare additional dependencies. For now, it has no effect.`

dxp/react_rendering.md
izaera: :shrug:

jbalsas: lovingit

dxp/react_rendering.md
izaera: Subordination is what distinguishes real writers from mere amateurs... :unamused:

izaera: Done.

dxp/react_rendering.md
izaera: makes sense.

izaera:
I'm merging this.

Issue #192, opened 2020-09-04
Document when to use Liferay.publish, Liferay.provide and Liferay.<myCustomFunctionallity/Property>


diegonvs:
## What is your question?

There are three ways to expose a thing on global `Liferay` object, where I don't know for sure when to use a method or not

wincent:
I don't recall what `Liferay.publish` does, but about the other two:

- You should never use `Liferay.provide` (it's AUI and we've actually done a lot of work removing usages; for example, look at the "Avoid Liferay.provide" tasks in [LPS-115885](https://issues.liferay.com/browse/LPS-115885)).
- You should only stick things on the `Liferay` global as a last resort; instead, encourage people to import what they need with `import {thing} from 'frontend-js-web'` (that's the short answer; there is a longer one).

jbalsas:
Just for reference, and since yuilibrary.com is no more 😢 

**[publish](https://github.com/liferay/yui3/blob/master/src/event-custom/js/event-target.js#L494-L496)**

> Creates a new custom event of the specified type.  If a custom event by that name already exists, it will not be re-created.  In either case the custom event is returned.

So that's deep YUI event system. Nothing we want to worry about now. 

As for the rest, @wincent put it very clearly... those takes might be the ones worth documenting 😉

wincent:
GitHub seems to have dropped the transfer job on the floor 😢

Most of the comments did not make it over to the transferred issue (https://github.com/liferay/liferay-frontend-projects/issues/340).

Oh well, going to close this anyway because I want to clean this whole repo of issues before requesting that it be archived.

PR #198, opened 2020-09-09
docs: Adds Liferay's Global Event Handler docs

### LINKED ISSUE #182: Document DXP global event emitter(Liferay.on, Liferay.after, Liferay.fire...)
Document DXP global event emitter

* Research and document DXP's ESB like Liferay.on, Liferay.after, Liferay.fire, Liferay.detach and others
* When to use it and not to use it

diegonvs:
Closes #182

### Questions

There are a lot of usages with getOpener() function. Can I add it as a best practice for modals? am I correct when mentioning this function is used on the iframes to get the root JS context?

Could I document somewhere that this event system will be rewritten in ES6 for migrating from AUI/YUI deps? Do we have an existing issue for it?


dxp/liferay_global_event_handler.md
```diff
@@ -0,0 +1,213 @@
...
+```js
+import {ns} from 'frontend-js-web'; // or simply Liferay.Util.ns
+
+Liferay.fire(ns(portletNamespace,'CLOSE_MODAL'), {
+	id: 'myId'
+})
+
+// when not portletNamespace available
+
+Liferay.fire('commerce:CLOSE_MODAL'), {
```
julien, line 210: looks like you have an extra `)` here, which would probably cause a syntax error

dxp/liferay_global_event_handler.md
```diff
@@ -0,0 +1,213 @@
...
+##### Code example:
+
+```js
+Liferay.fire('showNavigationMenu', mapHover);
+```
+
+## Best Practices
+
+The best practice number #1 is using Liferay’s namespace utilities(Liferay.Util.ns) when possible. Like `Liferay.Util.ns(myWidgetNamespace, 'nameOfMyCustomEvent')`. For scoping the global name of the event. It will prevent you for listening to generic events which can be fired in different widgets that you may not be responsible.
+
```
julien, line 188: I think you should prefer importing the `ns` function rather than using the global just like you did in your example below (I know the global exists)

Instead of saying "(Liferay.Util.ns)" maybe you could add a link to the [`ns`](https://github.com/liferay/liferay-portal/blob/6a82184b83c93b770e7518d93bbd72d489f06908/modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/liferay/util/ns.es.js#L33) function


```import {ns} from 'frontend-js-web';``` > ```Liferay.Util.ns()```

File: dxp/liferay_global_event_handler.md [hunk 3]
```diff
@@ -0,0 +1,213 @@
...
+Liferay.on('allPortletsReady', function (event) {
+	document.getElementById('<%= HtmlUtil.escape(scroll) %>').scrollIntoView();
+});
+```
+
+### `Liferay.fire`
+
+Description from: [yui3/event-target.js at master · liferay/yui3 · GitHub](https://github.com/liferay/yui3/blob/master/src/event-custom/js/event-target.js#L723)
+
```
julien, line 162: I would change this `[yui3/event-target.js at master · liferay/yui3 · GitHub]` because I find it confusing for the reader: are we talking about yui3? liferay/yui3 or GitHub?

dxp/liferay_global_event_handler.md
```diff
@@ -0,0 +1,213 @@
...
+##### Code Example
+
+```js
+var onDestroyPortlet = function () {
+	delegateHandler.removeListener();
+
+	Liferay.detach('destroyPortlet', onDestroyPortlet);
+};
+
+Liferay.on('destroyPortlet', onDestroyPortlet);
```
julien, line 130: Not sure if this is needed but I think we want to use es6 in code examples; and we prefer function declarations over function expressions - (@wincent will confirm if I'm wrong)

dxp/liferay_global_event_handler.md
```diff
@@ -0,0 +1,213 @@
...
+`Liferay.publish`
+`Liferay.fire`
+
+### `Liferay.on`
+
+Description from [yui3/event-target.js at master · liferay/yui3 · GitHub](https://github.com/liferay/yui3/blob/master/src/event-custom/js/event-target.js#L217):
+
+This function subscribes a callback function to a custom event fired by this object or from an object that bubbles its events to this object.
+
+Callback functions for events published with `emitFacade = true` will receive an `EventFacade` as the first argument (typically named “e”). These callbacks can then call `e.preventDefault()` to disable the behaviour published to that event’s `defaultFn` . See the `EventFacade` API for all available properties and methods. Subscribers to non- `emitFacade` events will receive the arguments passed to `fire()` after the event name.
```
julien, line 29: > (typically named “e”)

That's probably still the case in DXP, but I think we want people to name that `event` because we don't advise using abbreviations

julien: > See the `EventFacade` API

Maybe we could add a link to that

dxp/liferay_global_event_handler.md
```diff
@@ -0,0 +1,213 @@
...
+
+To subscribe to multiple events at once, pass an object as the first argument, where the key:value pairs correspond to the eventName:callback, or pass an array of event names as the first argument to subscribe to all listed events with the same callback.
+
+Returning `false` from a callback is supported as an alternative to calling `e.preventDefault(); e.stopPropagation();` . However, it is recommended to use the event methods whenever possible.
```
julien, line 33: Good point

dxp/liferay_global_event_handler.md
```diff
@@ -0,0 +1,213 @@
...
+
+##### Code example:
+
+```js
+var onDestroyPortlet = function () {
+	Liferay.detach('messagePosted', onMessagePosted);
+	Liferay.detach('destroyPortlet', onDestroyPortlet);
+};
+
+Liferay.on('destroyPortlet', onDestroyPortlet);
```
julien, line 43: I think you should use  a function declaration here, that will let you remove that `var`

dxp/liferay_global_event_handler.md
```diff
@@ -0,0 +1,213 @@
...
+Here is a list of available functions from Liferay’s global variable:
+
+`Liferay.on`
+`Liferay.once`
+`Liferay.onceAfter`
+`Liferay.after`
+`Liferay.before`
+`Liferay.detach`
+`Liferay.detachAll`
+`Liferay.publish`
```
wincent, line 20: I don't think we want to mention `.publish` at all, as @jbalsas [said here](https://github.com/liferay/liferay-frontend-guidelines/issues/192#issuecomment-688109206):

> So that's deep YUI event system. Nothing we want to worry about now.

dxp/liferay_global_event_handler.md
```diff
@@ -0,0 +1,213 @@
+# Liferay’s Global Event Handler
```
wincent, line 1: Technically, I think "event bus" is the more commonly used term to describe the entire event system. "Handler" just refers to a specific function that is notified of an event when it happens.

Separately, I see "smart quotes" in this file. Truly evil, just sayin'.

true-evil

dxp/liferay_global_event_handler.md
```diff
@@ -0,0 +1,213 @@
...
+
+Liferay’s event handler helps the developers for creating/listening/firing/removing custom events that can be listened in whole DXP which is rely on the YUI3 event target implementation. See the implementation [here](https://github.com/liferay/yui3/blob/master/src/event-custom/js/event-target.js#L120).
```
wincent, line 3: I don't think we should emphasize YUI itself, because it is an _implementation detail_ which may well change in the future, even as we maintain the same outward facing API. Still fine to link to it, but I just wouldn't _emphasize_ it so much. In other words instead of:

> rely on the YUI3 event target implementation. See the implementation [here](https://github.com/liferay/yui3/blob/master/src/event-custom/js/event-target.js#L120).

You could just link one of the words in the first sentence:

> [Liferay's global event bus implementation](https://github.com/liferay/yui3/blob/master/src/event-custom/js/event-target.js#L120) allows developers... (etc)

If you wanted to keep your initial wording, you would want to avoid linking to the word "here" anyway (bad for screen readers, because when the user asks for a list of links on the page, "here" doesn't convey any useful context to them); instead of:

> rely on the YUI3 event target implementation. See the implementation [here](https://github.com/liferay/yui3/blob/master/src/event-custom/js/event-target.js#L120).

you would do:

> rely on [the YUI3 event target implementation](https://github.com/liferay/yui3/blob/master/src/event-custom/js/event-target.js#L120).

Overall, maybe some more natural English phrasing for the whole thing would be:

> [Liferay's global event bus implementation](https://github.com/liferay/yui3/blob/master/src/event-custom/js/event-target.js#L120) allows developers to publish and listen to custom events from anywhere inside DXP.

(The bit about creating/removing is probably implicit.)

dxp/liferay_global_event_handler.md
```diff
@@ -0,0 +1,213 @@
...
+
+### First Considerations
+
+First, just consider using the global event handler if necessary. Try to use the default event listener/[callback pattern(if React)](https://reactjs.org/docs/lifting-state-up.html#lifting-state-up) when working. Just use the global event handler when trying to communicate between widgets or not accessible contexts(e.g iframes).
```
wincent, line 7: Typos: Missing space before `(` in two places.

> First, just consider using the global event handler if necessary.

It's not clear to me what this sentence means:

- It _sounds_ like using `document.addEventListener` (something which should be kept to a minimum anyway); or:
- You mean something else by "the global event handler", like "don't use the Liferay global event without first trying to use a more local solution.

Based on the context, I think you mean the second one, right?

> the default event listener

It's also not clear to me what you're referring to there.

I think this whole paragraph might be clearer overall as something like:

> In many situations you do not need an actually _global_ event system because a more local form of communication may suffice. For example, React components often [store shared state in a common ancestor](https://reactjs.org/docs/lifting-state-up.html), passing data _down_ the tree via props or context, and communicating _upwards_ using callbacks or by dispatching actions.
>
> The Liferay event system is useful to communicate between isolated contexts (eg. widgets, iframes).

dxp/liferay_global_event_handler.md
```diff
@@ -0,0 +1,213 @@
...
+
+### List of Available Functions
+
+Here is a list of available functions from Liferay’s global variable:
```
wincent, line 11: If you wanted to do the "less is more" thing, heading + sentence could be reduced to just heading:

```
### Functions
```

dxp/liferay_global_event_handler.md
```diff
@@ -0,0 +1,213 @@
...
+`Liferay.after`
+`Liferay.before`
+`Liferay.detach`
+`Liferay.detachAll`
+`Liferay.publish`
+`Liferay.fire`
+
+### `Liferay.on`
+
+Description from [yui3/event-target.js at master · liferay/yui3 · GitHub](https://github.com/liferay/yui3/blob/master/src/event-custom/js/event-target.js#L217):
```
wincent, line 25: I don't think you need to link this as you already did above; just describe the function like you are already.

dxp/liferay_global_event_handler.md
wincent: > but I think we want people to name that event because we don't advise using abbreviations

Yep: https://github.com/liferay/liferay-frontend-guidelines/blob/master/general/naming.md#avoid-abbreviations

dxp/liferay_global_event_handler.md
```diff
@@ -0,0 +1,213 @@
...
+
+This function subscribes a callback function to a custom event fired by this object or from an object that bubbles its events to this object.
+
+Callback functions for events published with `emitFacade = true` will receive an `EventFacade` as the first argument (typically named “e”). These callbacks can then call `e.preventDefault()` to disable the behaviour published to that event’s `defaultFn` . See the `EventFacade` API for all available properties and methods. Subscribers to non- `emitFacade` events will receive the arguments passed to `fire()` after the event name.
+
+To subscribe to multiple events at once, pass an object as the first argument, where the key:value pairs correspond to the eventName:callback, or pass an array of event names as the first argument to subscribe to all listed events with the same callback.
```
wincent, line 31: I don't think anybody is doing this in liferay-portal, so I wouldn't document it:

```
git grep 'Liferay.on\(\{'
```

... returns no hits.

dxp/liferay_global_event_handler.md
wincent: Don't forget `e` → `event` here.

dxp/liferay_global_event_handler.md
wincent: Given that you're not defining `onMessagePosted` here, it seems extraneous and distracting; I'd just remove it.

I'm also not sure I like the synthetic example here of a `destroyPortlet` listener that does nothing but detach itself (you would have used `once` if that's what what you really wanted to do).

An example like this one seems a bit more appropriate:

```js
Liferay.on('destroyPortlet', (event) => {
    // Perform clean-up here, access event.portletId etc...
});
```

dxp/liferay_global_event_handler.md
```diff
@@ -0,0 +1,213 @@
...
+};
+
+Liferay.on('destroyPortlet', onDestroyPortlet);
+```
+
+### `Liferay.once`:
+
+Description from [yui3/event-target.js at master · liferay/yui3 · GitHub](https://github.com/liferay/yui3/blob/master/src/event-custom/js/event-target.js#L136):
+
+Listen to a custom event hosted by the global event handler one time.
```
wincent, line 50: I'd delete these two lines.

dxp/liferay_global_event_handler.md
```diff
@@ -0,0 +1,213 @@
...
+Liferay.once(this.ns('formReady'), (event) => {
+	const form = Liferay.Form.get(event.formName);
+
+	const field = form.formValidator.getField(this.ns('emailAddress'));
+
+	if (field) {
+		if (this.viewValidDomainsURL) {
+			this.addFieldMessage_(field);
+		}
+	}
```
wincent, line 66: This example is super interesting if you want to learn about `Liferay.Form`, but irrelevant for learning what `Liferay.once` does. I think you could just blow it all away:

```js
Liferay.once('someEvent', (event) => {
    // Called only once, no need to call `detach()`.
});
```

dxp/liferay_global_event_handler.md
wincent: Sorry, not a good example that one because I'm omitting the possibility of clean-up.

```js
function destroyPortlet(event) {
    // Perform clean-up here, access event.portletId etc...
}

Liferay.on('destroyPortlet', destroyPortlet);
```

wincent (commented):
I just read the first bit and I don't want to spam it with too many comments about "English style tweaks", so I'm going to stop for now: I figure we can come back and copy edit for those at the end before merging.

dxp/liferay_global_event_handler.md
diegonvs: 👀

diegonvs: > I think "event bus" is the more commonly used term to describe the entire event system. "Handler" just refers to a specific function that is notified of an event when it happens.

Yes, "handler" is just a part of it. I'll change :)

diegonvs: When searching about DXP "event bus" I found some links that mentions "IPC(Inter Portlet Communication)" instead. Should I update and use Event Bus or IPC? What do you think? @wincent @jbalsas @david-truong

See:
- http://www.liferaysavvy.com/2014/01/liferay-client-side-inter-portlet.html
- https://liferay.dev/forums/-/message_boards/message/38984904

dxp/liferay_global_event_handler.md
diegonvs: >  Still fine to link to it, but I just wouldn't emphasize it so much.

Makes sense. I forgot to update PR's questions section for asking if we have a ticket for this kinda modernization. I tried now searching for things with `FI_Technical_Debt` label and I didn't found :/

/cc @jbalsas

dxp/liferay_global_event_handler.md
diegonvs: > Based on the context, I think you mean the second one, right?

Yep! 

> I think this whole paragraph might be clearer overall as something like:

Thanks for the wordsmith 💖

dxp/liferay_global_event_handler.md
diegonvs: > Don't forget e → event here.

I'll fix it!

dxp/liferay_global_event_handler.md
diegonvs: > I think you should use a function declaration here, that will let you remove that var

I just copied and paste a JSP usage. Would be good point the file, isn't? I think it will look more practical :)

dxp/liferay_global_event_handler.md
diegonvs: Makes sense. However, I notice we're using on our codebase(AFAIK on frontend-js-web) and It may look weird for people who don't touch on AUI/YUI things read it.

dxp/liferay_global_event_handler.md
diegonvs: It looks more simpler, yeh? I'll update :)

dxp/liferay_global_event_handler.md
diegonvs: I pointed to our yui3 fork because I was guessing we have some fixes and behavior changed as `liferay-ckeditor` but it's okay just point to the original repo :)

dxp/liferay_global_event_handler.md
diegonvs: > import {ns} from 'frontend-js-web'; > Liferay.Util.ns()

Yes, I was providing the example using global functions because We have a lot of usages on JSPs but I can empathize with it.

dxp/liferay_global_event_handler.md
diegonvs: Thanks <3

dxp/liferay_global_event_handler.md
diegonvs: Now I got it, people weren't using the configs parameter

dxp/liferay_global_event_handler.md
diegonvs: 👍 Since we're avoiding abbreviations on DXP code makes sense remove it :)

jbalsas:
> There are a lot of usages with getOpener() function. Can I add it as a best practice for modals? am I correct when mentioning this function is used on the iframes to get the root JS context?

We are not trying to write best practices now. Just understand what we have and how it works and is being used so we can use that to improve the platform in the future.

dxp/liferay_global_event_handler.md
jbalsas: IPC is something completely different, and DXP has indeed a Message (Event) Bus but that refers to a backend event bus. 

This is meant for internal consumption so we shouldn't need to get into small distinctions like that. Our target audience (internal liferay frontend developers) won't be confused by those terms (if they even know what they mean).

dxp/liferay_global_event_handler.md
jbalsas: Me and @dsanz will be creating tickets as we build our roadmap and backlog. You knowing this can help us fill in the gaps in the next weeks if you don't see it around.

dxp/liferay_global_event_handler.md
jbalsas: Let's add those recommendations and push, @diegonvs! The `widgets` addition to the list of "isolated contexts" is particularly interesting because this pattern is more widespread than if it were just to communicate iframes.

dxp/liferay_global_event_handler.md
jbalsas: I would simplify this one, it doesn't even have to be real code:

```javascript
function doSomething(event) {
    // Do something here when 'something' happens 
}

Liferay.on('something', doSomething);
```

Then show the `.detach` method down there so it does not get mixed up with the `.on` explanation.

dxp/liferay_global_event_handler.md
jbalsas: Agreed, no need to show unrelated code 😉

dxp/liferay_global_event_handler.md
```diff
@@ -0,0 +1,213 @@
...
+Subscribe to a custom event hosted by the global event handler.
+
+The supplied callback will execute after any listeners add via the subscribe method, and after the default function, if configured for the event, has executed.
+
+##### Code Example:
+
+```js
+Liferay.after('commerce:productAddedToCart', function (event) {
+	Liferay.Portlet.refresh('#p_p_id<portlet:namespace />');
+});
```
jbalsas, line 106: Simplify the example ;)

dxp/liferay_global_event_handler.md
```diff
@@ -0,0 +1,213 @@
...
+Liferay.on('destroyPortlet', onDestroyPortlet);
+```
+
+### `Liferay.detachAll`
+
+Description from: [yui3/event-target.js at master · liferay/yui3 · GitHub](https://github.com/liferay/yui3/blob/master/src/event-custom/js/event-target.js#L476)
+
+Removes all listeners from the specified event. If the event type is not specified, all listeners from all hosted custom events will be removed.
+
+### `Liferay.publish`
```
jbalsas, line 139: Let's just remove this for now 😉

dxp/liferay_global_event_handler.md
wincent: Yeah, I was just talking about the functionality described on line 31; which nobody is using (unless it's hidden very deeply... Prettier should always keep the curly on the same line after `Liferay.on(`, which is why I did the grep like that).

dxp/liferay_global_event_handler.md
```diff
@@ -0,0 +1,213 @@
...
+```js
+Liferay.onceAfter(
+    '<%= portletDisplay.getId() %>:messagePosted',
+    function (event) {
+        <%= randomNamespace %>onMessagePosted(
+            response,
+            refreshPage
+        );
+    }
+);
```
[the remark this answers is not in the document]
diegonvs: I'll update the examples following your guidance :)

dxp/liferay_global_event_handler.md
diegonvs: > Liferay's global event bus implementation allows developers to publish and listen to custom events from anywhere inside DXP.

Looks fine :)

dxp/liferay_global_event_handler.md
diegonvs: Done

dxp/liferay_global_event_handler.md
diegonvs: Removed!

dxp/liferay_global_event_handler.md
diegonvs: Updated!

dxp/liferay_global_event_handler.md
diegonvs: it was turned: Liferay’s Global Event Bus 🎉

dxp/liferay_global_event_handler.md
diegonvs: Removed, as the original path was mentioned on the document heading. As discussed on https://github.com/liferay/liferay-frontend-guidelines/pull/198#discussion_r485456671

dxp/liferay_global_event_handler.md
diegonvs: Removed, as the original path was mentioned on the document heading. As discussed on https://github.com/liferay/liferay-frontend-guidelines/pull/198#discussion_r485456671

dxp/liferay_global_event_handler.md
diegonvs: Updated!

dxp/liferay_global_event_bus.md
```diff
@@ -0,0 +1,170 @@
+# Liferay’s Global Event Bus
+
+[Liferay's global event bus](https://github.com/liferay/yui3/blob/master/src/event-custom/js/event-target.js#L120) implementation allows developers to publish and listen to custom events from anywhere inside DXP.
+
+### First Considerations
+
+In many situations you do not need an actually global event system because a more local form of communication may suffice. For example, React components often [store shared state in a common ancestor](https://reactjs.org/docs/lifting-state-up.html), passing data down the tree via props or context, and communicating upwards using callbacks or by dispatching actions.
```
wincent, line 7: tiny nit: "actually" → "actual"

wincent:
Might want to squash this before merging. Also "refactor" is for code changes; these are all "docs" changes.

Screenshot 2020-09-11 -085952-dtiwRGKo@2x

diegonvs:
> Might want to squash this before merging. Also "refactor" is for code changes; these are all "docs" changes.

Initially, I used the type `docs` for whole commits then I thought and I look like a refactoring and I rebased it 🤦

diegonvs:
I'll squash and push force :)

diegonvs:
image

 @wincent

wincent:
Nice! Don't forget to address [this comment](https://github.com/liferay/liferay-frontend-guidelines/pull/198/files#r486813432) and take this out of draft status.

dxp/liferay_global_event_bus.md
diegonvs: done!

diegonvs:
Updated! 

```git
commit e7e95394db11be92fab5fc111106b99b251c91f4 (HEAD -> global-event-handler, origin/global-event-handler)
Author: Diego Nascimento <diego.nascimento@liferay.com>
Date:   Wed Sep 9 17:54:23 2020 -0300

    docs: Adds review feedback

    It does:

    * Prefer functions declarations instead of functions expressions
    * Since it was already mentioned in the beginning of the file doesn't make sense repeat the same path for each function
    * Points to ns definition instead of quoting the global variable shortcut
    * Format headings, adds a proper summary and transforms `List of Available Functions` -> `Functions`
    * Renames from "Liferay's Global Event Handler" to "Liferay's Global Event Bus"
    * Update `Liferay.detach` code example as proposed on https://github.com/liferay/liferay-frontend-guidelines/pull/198#discussion_r486349988
    * Avoids abbreviations
    * Simplifies code examples and use a modern syntax for JS
    * Simplifies the `Liferay.onceAfter` code example
    * Simplifies `Liferay.once` code example
    * Remove the first argument as an object on `Liferay.on` since there aren't any usages on DXP

    When running `git grep 'Liferay.on\(\{'` on `liferay-portal` root wasn't found any references
    * `actually` -> `actual`

commit adfec585c149a160424824d2fe085545e74a63e6
Author: Diego Nascimento <diego.nascimento@liferay.com>
Date:   Tue Sep 8 21:33:28 2020 -0300

    docs: Adds Liferay's Global Event Handler docs
```

wincent:
Looks good @diegonvs. I applied some tweaks for things that slipped through the review (separate commits so you can see what I changed), and rebased it too. This doesn't need to be perfect, so I'll merge it and we can follow-up with any additions or changes as we see fit.

PR #199, opened 2020-09-09
docs: Document Session API

kresimir-coko:
First time making a PR here, so let me know if I did something wrong! IDK how to check if the markdown is good, and hopefully the commit & PR names are good.

### DIFF
--- a/general/session.md
+++ b/general/session.md
@@ -1,0 +1,26 @@
+# Session API
+
+## Workflow
+
+### Getting
+
+When getting a value stored in Session, we are deserializing it by removing the `serialize://` string from the data we retrieve. After the data is deserialized we get access to the data requested, stripped of any serialization.
+
+#### Needs clarifying
+
+-   Even though we are using POST we are sending a `'cmd': 'get'`, which might be revealing the underlying architecture of how Liferay handles fetch requests.
+-   Why are we using a POST method in a getter?
+
+### Setting
+
+When setting values to Session, we have to serialize them first so that it is served in a proper way that the endpoint expects. After the value has been serialized, we append it to FormData which then get's sent to the `session_click` endpoint to be consumed at a later time.
+
+## Components used by the Session API
+
+### URL
+
+The base URL used by the Session API is `http://localhost:8080/c/portal/session_click` where `http://localhost:8080/` and `c/` are different depending on the server. This URL is appended by FormData, which consists of `'cmd': 'get'` (when using `getSessionValue`) or `'cmd': 'set'` (when using `setSessionValue`), and `'p_auth: Liferay.authToken`
+
+### Serialization and deserialization
+
+The data in APIs like Session needs to be properly formatted, either before being consumed or after it's been consumed, this process is called `serialization`. In the Session API we use `serialize://` which gets prepended to the values we are setting.
--- a/general/session.md
+++ b/general/session.md
@@ -1,26 +1,16 @@
 # Session API
 
-## Workflow
+The Session API provides a mechanism for associating persisted key/value pairs with a user as they navigate between pages.
 
-### Getting
+It consists of two methods exported from `frontend-js-web`:
 
-When getting a value stored in Session, we are deserializing it by removing the `serialize://` string from the data we retrieve. After the data is deserialized we get access to the data requested, stripped of any serialization.
+-   [getSessionValue(key)](https://github.com/liferay/liferay-portal/blob/2d2e8ae74e9e7b33a07328277da64f54db576a04/modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/liferay/util/session.es.js#L38-L63)
+-   [setSessionValue(key, Value)](https://github.com/liferay/liferay-portal/blob/2d2e8ae74e9e7b33a07328277da64f54db576a04/modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/liferay/util/session.es.js#L65-L85)
 
-#### Needs clarifying
+## Setting
 
--   Even though we are using POST we are sending a `'cmd': 'get'`, which might be revealing the underlying architecture of how Liferay handles fetch requests.
--   Why are we using a POST method in a getter?
+`setSessionValue(key, data)` makes a request to the server to associate `data` with `key` in the user's session. If `data` is a non-falsy object, it will be encoded with `JSON.stringify()` and prefixed with a special marker, `serialize://`, to flag it for deserialization on later retrieval. Otherwise, it gets transmitted as-is.
 
-### Setting
+## Getting
 
-When setting values to Session, we have to serialize them first so that it is served in a proper way that the endpoint expects. After the value has been serialized, we append it to FormData which then get's sent to the `session_click` endpoint to be consumed at a later time.
-
-## Components used by the Session API
-
-### URL
-
-The base URL used by the Session API is `http://localhost:8080/c/portal/session_click` where `http://localhost:8080/` and `c/` are different depending on the server. This URL is appended by FormData, which consists of `'cmd': 'get'` (when using `getSessionValue`) or `'cmd': 'set'` (when using `setSessionValue`), and `'p_auth: Liferay.authToken`
-
-### Serialization and deserialization
-
-The data in APIs like Session needs to be properly formatted, either before being consumed or after it's been consumed, this process is called `serialization`. In the Session API we use `serialize://` which gets prepended to the values we are setting.
+`getSessionValue(key)` retrieves string data from the user's session by making a fetch request to the server at a [`portal/session_click` endpoint](https://github.com/liferay/liferay-portal/blob/2d2e8ae74e9e7b33a07328277da64f54db576a04/modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/liferay/util/session.es.js#L34-L36). If the retrieved data begins with the `serialize://` prefix, it is stripped off and the remaining text is parsed as JSON using `JSON.parse()`. Otherwise, the data string is returned as-is.


wincent:
> IDK how to check if the markdown is good, and hopefully the commit & PR names are good.

`yarn format` will make sure Markdown is well formatted and you can check the visuals here:

Screenshot 2020-09-09 -164739-YONovpTE@2x

(full view)

Or here:

Screenshot 2020-09-09 -164829-SebHvpfy@2x

("diff" view)

As for:

>  the commit & PR names are good

Like in the Clay repo, the Semantic PR bot will tell you if the format is wrong, although it has no opinions on what makes a "good" message. 😂

general/web_apis.md
```diff
@@ -0,0 +1,40 @@
+# Best practices when using Web APIs
+
+## Local Storage
+
+When writing implementations using `localStorage` first we must understand that `localStorage` is prone to evaluating as `undefined`.
```
wincent, line 5: "prone" makes it sounds like it happens all the time. I actually think it's pretty rare that somebody would turn it off in their prefs. Maybe just say "may evaluate as `undefined`)".

general/web_apis.md
```diff
@@ -0,0 +1,40 @@
...
+
+This [article](https://michalzalecki.com/why-using-localStorage-directly-is-a-bad-idea/) is a good starting point explaining most common pitfalls when using `localStorage`.
+
+When using `localStorage` we recommend first testing if it exists in the current `window` object. We do this by attempting to set an item to it: `localStorage.setItem()` & `localStorage.removeItem()`. If it successfully sets and then removes the item, that means that `localStorage` is defined, and therefore it can be used in the app.
```
wincent, line 9: "&" → "and"

wincent (commented):
Looks fine to me, although I think we _shouldn't_ actually be in the business of writing `localStorage` wrappers. The two places we'll ever need to do this are probably the Clay repo (done) and DXP (where we may already have some abstraction around it; I don't know). 🤷 

Nevertheless, feel free to merge.

jbalsas:
> Looks fine to me, although I think we shouldn't actually be in the business of writing localStorage wrappers. The two places we'll ever need to do this are probably the Clay repo (done) and DXP (where we may already have some abstraction around it; I don't know). 🤷

I agree with this.

I'd much rather see a document describing what developers can currently use for session-based storage explaining how it works. 

The whole goal of this exercise is to gain knowledge about our architecture so we can make improvements in the future with a broader vision of the impact that might have.

kresimir-coko:
> I'd much rather see a document describing what developers can currently use for session-based storage explaining how it works.

@jbalsas If I understand that correctly, I should change the approach and write how people can consume the Clay wrapper, and the abstraction for it that we have in DXP, instead of writing the wrapper itself? And explain what using these "helpers/wrappers" is useful?

jbalsas:
> @jbalsas If I understand that correctly, I should change the approach and write how people can consume the Clay wrapper, and the abstraction for it that we have in DXP, instead of writing the wrapper itself? And explain what using these "helpers/wrappers" is useful?

What I'd like to see is you research and find what mechanisms **exist already in DXP** and are **already being used**. Then learn how they work and write about them. 

With that knowledge, then we can start thinking about how to bring new things to the table.

jbalsas:
If you can't find any in a reasonable amount of time, you may ask and we may provide some hints 😉

kresimir-coko:
Hey @jbalsas can you let me know if I'm on the right track? I sent a commit with bullet points, because I want to get some quick feedback before expanding those bullet points into paragraphs.

jbalsas:
Hey @kresimir-coko!

Getting slightly closer! While there's some good information here (like who's using `localStorage` directly, this is still pretty much describing something that others are using and that may or may not be a good pattern.

> Rest of DXP doesn't have any specific ways of using localStorage or sessionStorage, those places could probably use with some kind of sanity check before using it.

The fact that at the end you say this means you didn't really find "mechanisms exist already in DXP". Probably because you're looking specifically at `localStorage` as a specific API and not at it from a functional perspective.

What I had in mind was:
- [getSessionValue](https://github.com/liferay-frontend/liferay-portal/blob/ee3868adec74d31e8bd55fcaee36cf3d5c546fc5/modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/liferay/util/session.es.js#L44)
- [setSessionValue](https://github.com/liferay-frontend/liferay-portal/blob/ee3868adec74d31e8bd55fcaee36cf3d5c546fc5/modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/liferay/util/session.es.js#L72)

That's as far as I know a similar (yet different) concept that we have and use in DXP. Could you try taking it from here? See the tests, check the URL calls it makes, where it stores the values, how it retrieves them... learning and describing that architecture is what can help us make smarter design decisions later 😉

kresimir-coko:
I see why I didn't find this file, I was searching `localStorage`, `sessionStorage` and `storage`, but not `session`, okay thanks for giving me some clues!

kresimir-coko:
I've pushed a commit with some bullet points inspired by your comment @jbalsas and filling them up with some basic info from the Session utility. This is by no means done, just putting it out there as I'm done for the day.

jbalsas:
> I've pushed a commit with some bullet points inspired by your comment @jbalsas and filling them up with some basic info from the Session utility. This is by no means done, just putting it out there as I'm done for the day.

Looks to me like the good approach! 

And yes, I would scratch the other thing about localStorage for now. At least from here, but not from your mind 😉

kresimir-coko:
I've expanded upon the bullet points and written short paragraphs split into a few categories.

There are a couple bullet points in which I basically ask for clarification or straight up answers, I'll write them here too:
- Even though we are using POST we are sending a `'cmd': 'get'`, which might be revealing the underlying architecture of how Liferay handles fetch requests. 
- Why are we using a POST method in a getter?

The 2 bullet points are referencing the same getter method, but I'm not sure about the underlying architecture of handling fetch requests with the `'cmd': 'get'`, and I'm unsure why we're using a POST method inside that same getter.

kresimir-coko:
@jbalsas in case you didn't notice it because I didn't initially ping you this requires reviewing and some clarification as I mentioned in the comment above.

general/session.md
```diff
@@ -0,0 +1,26 @@
+# Session API
```
wincent, line 1: Can we lead off here with a link to the API, and a statement of what it is actually for?

---

The Session API provides a mechanism for associating persisted key/value pairs with a user as they navigate between pages.

It consists of two methods exported from `frontend-js-web`:

- [`getSessionValue(key)`](https://github.com/liferay/liferay-portal/blob/2d2e8ae74e9e7b33a07328277da64f54db576a04/modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/liferay/util/session.es.js#L38-L63)
- [`setSessionValue(key, Value)`](https://github.com/liferay/liferay-portal/blob/2d2e8ae74e9e7b33a07328277da64f54db576a04/modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/liferay/util/session.es.js#L65-L85)

general/session.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Workflow
+
+### Getting
+
+When getting a value stored in Session, we are deserializing it by removing the `serialize://` string from the data we retrieve. After the data is deserialized we get access to the data requested, stripped of any serialization.
```
wincent, line 7: This seems incorrect. Maybe something like this instead?

---

`getSessionValue(key)` retrieves string data from the user's session by making a `fetch` request to the server at [a `portal/session_click` endpoint](https://github.com/liferay/liferay-portal/blob/2d2e8ae74e9e7b33a07328277da64f54db576a04/modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/liferay/util/session.es.js#L34-L36). If the retrieved data begins with the `serialize://` prefix, it is stripped off and the remaining text is parsed as JSON using `JSON.parse()`. Otherwise, the data string is returned as-is.

general/session.md
```diff
@@ -0,0 +1,26 @@
...
+
+#### Needs clarifying
+
+-   Even though we are using POST we are sending a `'cmd': 'get'`, which might be revealing the underlying architecture of how Liferay handles fetch requests.
+-   Why are we using a POST method in a getter?
```
wincent, line 12: Probably doesn't need to be addressed in this documentation, which is mostly about usage. The point of having a `getSessionValue()` method is to abstract over these details so that the developer doesn't have to worry about them too much.

Talking about `serialize://` is getting into implementation details, but I think it's an acceptable level of detail to go into, especially if we reorder the sections so that `setSessionValue()` comes first and we explain how the prefix gets put in there in the first place.

general/session.md
```diff
@@ -0,0 +1,26 @@
...
+
+### Setting
+
+When setting values to Session, we have to serialize them first so that it is served in a proper way that the endpoint expects. After the value has been serialized, we append it to FormData which then get's sent to the `session_click` endpoint to be consumed at a later time.
```
wincent, line 16: Doesn't seem to describe what is actually happening. Maybe something like?

---

`setSessionValue(key, data)` makes a request to the server to associate `data` with `key` in the user's session. If `data` is a non-falsy object, it will be encoded with `JSON.stringify()` and prefixed with a special marker, `serialize://`, to flag it for deserialization on later retrieval. Otherwise, it gets transmitted as-is.

general/session.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Components used by the Session API
+
+### URL
+
+The base URL used by the Session API is `http://localhost:8080/c/portal/session_click` where `http://localhost:8080/` and `c/` are different depending on the server. This URL is appended by FormData, which consists of `'cmd': 'get'` (when using `getSessionValue`) or `'cmd': 'set'` (when using `setSessionValue`), and `'p_auth: Liferay.authToken`
+
+### Serialization and deserialization
+
+The data in APIs like Session needs to be properly formatted, either before being consumed or after it's been consumed, this process is called `serialization`. In the Session API we use `serialize://` which gets prepended to the values we are setting.
```
wincent, line 26: At this point, I think it's too late to say this, so I'd just omit this section and roll the info up into the sections above like in my proposed edits.

general/session.md
```diff
@@ -0,0 +1,26 @@
...
+
+### Setting
+
+When setting values to Session, we have to serialize them first so that it is served in a proper way that the endpoint expects. After the value has been serialized, we append it to FormData which then get's sent to the `session_click` endpoint to be consumed at a later time.
+
+## Components used by the Session API
+
+### URL
+
+The base URL used by the Session API is `http://localhost:8080/c/portal/session_click` where `http://localhost:8080/` and `c/` are different depending on the server. This URL is appended by FormData, which consists of `'cmd': 'get'` (when using `getSessionValue`) or `'cmd': 'set'` (when using `setSessionValue`), and `'p_auth: Liferay.authToken`
```
wincent, line 22: I don't think any of this matters. If somebody is interested about it, they can look at the source. The purpose of these docs is above all else to:

1. Provide a guide for usage (probably pretty trivial if we're talking about methods which take params `key` and `key, data`, respectively).
2. Give a high-level sense of what this API is for, conceptually.

I don't think translating the implementation from JS to English is really worth the time and effort though (at best, it's work you didn't really need to do because you could have linked to the functions which are both very short, and at worst, it's inaccurate).

wincent (changes requested):
@kresimir-coko: I have some suggested changes, so would you mind making those and we can merge this? I expect the result will be pretty short — this is not the kind of thing that we need to speak very extensively on.

general/session.md
kresimir-coko: This was mainly for myself, so I can  understand the API better in order to either explain it better before merging, so if you know why we're using `POST` instead of `GET`, and how the whole `'cmd': 'get'` thingy functions it would be nice 😄

kresimir-coko:
I think that the way I described the API was a bit different than how it was supposed to be explained, I feel like I understood the way it works, just didn't write with the proper perspective.

Thanks for essentially writing this for me in the end @wincent haha, it was supposed to be `localStorage` and `sessionStorage` initially, I didn't even know `Session` existed before I started writing this.

I pushed a commit that has the changes you requested.

jbalsas:
> I didn't even know Session existed before I started writing this.

That's one of the goals of this whole exercise! 😉

wincent (approved):
For your next challenge, my young padawan, you must document `SimpleBeanFactoryAwareAspectInstanceFactory`, if you can even find it...

https://media.giphy.com/media/l3fZG8Gsb7bs9MZpK/giphy.gif

PR #200, opened 2020-09-10
docs: document resource serving infrastructure

izaera:

PR #201, opened 2020-09-11
docs: JavaScript minification in DXP

### LINKED ISSUE #177: DXP Minification Architecture
Document how does **DXP support Minification of static resources (JS, CSS...)**.

Some references:
- [frontend-js-minifier](https://github.com/liferay/liferay-portal/tree/28d106459e63294b4ea7b70bc4235b2b570d6de6/modules/apps/frontend-js/frontend-js-minifier)
- [MinifierUtil](https://github.com/liferay/liferay-portal/blob/28d106459e63294b4ea7b70bc4235b2b570d6de6/portal-impl/src/com/liferay/portal/minifier/MinifierUtil.java)
- [AggregateFilter](https://github.com/liferay/liferay-portal/blob/28d106459e63294b4ea7b70bc4235b2b570d6de6/portal-impl/src/com/liferay/portal/servlet/filters/aggregate/AggregateFilter.java)
- [ComboServlet](https://github.com/liferay/liferay-portal/blob/28d106459e63294b4ea7b70bc4235b2b570d6de6/portal-impl/src/com/liferay/portal/servlet/ComboServlet.java)
- [StripFilter](https://github.com/liferay/liferay-portal/blob/28d106459e63294b4ea7b70bc4235b2b570d6de6/portal-impl/src/com/liferay/portal/servlet/filters/strip/StripFilter.java)
- [JSLoaderConfigServlet](https://github.com/liferay/liferay-portal/blob/28d106459e63294b4ea7b70bc4235b2b570d6de6/modules/apps/frontend-js/frontend-js-loader-modules-extender/src/main/java/com/liferay/frontend/js/loader/modules/extender/internal/servlet/JSLoaderConfigServlet.java)
- [JSBundleConfigServlet](https://github.com/liferay/liferay-portal/blob/28d106459e63294b4ea7b70bc4235b2b570d6de6/modules/apps/frontend-js/frontend-js-bundle-config-extender/src/main/java/com/liferay/frontend/js/bundle/config/extender/internal/JSBundleConfigServlet.java)

izaera:
Document infrastructure for JavaScript minification in DXP .

See #177.

> This CI build is failing because #200 has not been merged and references files in it.

### DIFF
--- a/dxp/javascript_minification.md
+++ b/dxp/javascript_minification.md
@@ -12,6 +12,8 @@ These three pieces form the basic infrastructure available in DXP to minify Java
 
 -   [`frontend-js-minifier`](https://github.com/liferay/liferay-portal/blob/3c6bbac29b64769b665e37798c7f51d4e45b6ff0/portal-impl/src/com/liferay/portal/minifier/MinifierUtil.java): this project contains the two default minifiers that DXP ships out-of-the-box (Google and Yahoo). See the two following sections for more details.
 
+> Note that, even though, `JavaScriptMinifier` and `MinifierUtil` are inside `portal-impl` to maintain backward compatibility, the implementation of the minifiers has been extracted to an OSGi module (`frontend-js-minifier`) for modularity.
+
 ### `GoogleJavaScriptMinifier`
 
 This is the default DXP minifier (because it is deployed with a higher service ranking than the other ones available).
--- a/dxp/javascript_minification.md
+++ b/dxp/javascript_minification.md
@@ -2,23 +2,23 @@
 
 This article describes DXP infrastructure to minify JavaScript source files. It is related to [Resource Serving in DXP](resource_serving.md).
 
-## MinifierUtil, JavaScriptMinifier and frontend-js-minifier
+## `MinifierUtil`, `JavaScriptMinifier` and `frontend-js-minifier`
 
-These three pieces conform the basic infrastructure available in DXP to minify JavaScript code.
+These three pieces form the basic infrastructure available in DXP to minify JavaScript code.
 
--   [JavaScriptMinifier](https://github.com/liferay/liferay-portal/blob/master/portal-impl/src/com/liferay/portal/minifier/JavaScriptMinifier.java): describes the interface that services to minify JavaScript must implement.
+-   [`JavaScriptMinifier`](https://github.com/liferay/liferay-portal/blob/3c6bbac29b64769b665e37798c7f51d4e45b6ff0/portal-impl/src/com/liferay/portal/minifier/JavaScriptMinifier.java): describes the interface that services to minify JavaScript must implement.
 
--   [MinifierUtil](https://github.com/liferay/liferay-portal/blob/master/portal-impl/src/com/liferay/portal/minifier/MinifierUtil.java): provides a way to retrieve the `JavaScriptMinifier` with highest service ranking currently deployed to DXP. Note that you can also obtain a reference to a `JavaScriptMinifier` (or all of them) using standard OSGi mechanisms.
+-   [`MinifierUtil`](https://github.com/liferay/liferay-portal/blob/3c6bbac29b64769b665e37798c7f51d4e45b6ff0/portal-impl/src/com/liferay/portal/minifier/MinifierUtil.java): provides a way to retrieve the `JavaScriptMinifier` with highest service ranking currently deployed to DXP. Note that you can also obtain a reference to a `JavaScriptMinifier` (or all of them) using standard OSGi mechanisms (for example: [`@Component`](https://docs.osgi.org/javadoc/r4v43/cmpn/org/osgi/service/component/annotations/Component.html) or [`ServiceTracker`](https://docs.osgi.org/javadoc/r6/core/org/osgi/util/tracker/ServiceTracker.html)).
 
--   [frontend-js-minifier](https://github.com/liferay/liferay-portal/tree/master/modules/apps/frontend-js/frontend-js-minifier): this project contains the two default minifiers that DXP ships out of the box (Google and Yahoo). See the two following sections for more details.
+-   [`frontend-js-minifier`](https://github.com/liferay/liferay-portal/blob/3c6bbac29b64769b665e37798c7f51d4e45b6ff0/portal-impl/src/com/liferay/portal/minifier/MinifierUtil.java): this project contains the two default minifiers that DXP ships out-of-the-box (Google and Yahoo). See the two following sections for more details.
 
-### GoogleJavaScriptMinifier
+### `GoogleJavaScriptMinifier`
 
 This is the default DXP minifier (because it is deployed with a higher service ranking than the other ones available).
 
-It uses [Google Closure Compiler](https://github.com/google/closure-compiler) under the hood. Because this is an _internal_ library from Google (thus unsupported and subject to big breaking changes) updating it has usually a significant cost so it is not done very often.
+It uses [Google Closure Compiler](https://github.com/google/closure-compiler) under the hood. Because this is an _internal_ library from Google (thus unsupported and subject to big breaking changes) updating it usually has a significant cost so it is not done very often.
 
-### YahooJavaScriptMinifier
+### `YahooJavaScriptMinifier`
 
 This uses [YUI Compressor](https://github.com/yui/yuicompressor) from Yahoo!, which is still active but unmaintained (except for some PRs that get merged from time to time). The last commit at the time of this writing (Sep 2020) is from May 2019.
 
@@ -28,13 +28,13 @@ Unless someone has a very good reason to use it, this shouldn't be the choice fo
 
 Currently, the following components in DXP use the minifier services.
 
-### StripFilter
+### `StripFilter`
 
 This is a filter used to minify inline JavaScript code contained in `<script>` tags and embedded in the HTML code.
 
 > It also does transformations for `<input>`, `<textarea>`, `<pre>` and `<style>` but that's out of the scope of this document.
 
-It filters any resource with its URL starting with (as specified in [liferay-web.xml](https://github.com/liferay/liferay-portal/blob/master/portal-web/docroot/WEB-INF/liferay-web.xml)):
+It filters any resource with its URL starting with (as specified in [`liferay-web.xml`](https://github.com/liferay/liferay-portal/blob/3c6bbac29b64769b665e37798c7f51d4e45b6ff0/portal-web/docroot/WEB-INF/liferay-web.xml)):
 
 -   `/c/...`
 -   `/group/...`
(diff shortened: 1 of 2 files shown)


dxp/javascript_minification.md
```diff
@@ -0,0 +1,86 @@
+# JavaScript Minification in DXP
+
+This article describes DXP infrastructure to minify JavaScript source files. It is related to [Resource Serving in DXP](resource_serving.md).
+
+## MinifierUtil, JavaScriptMinifier and frontend-js-minifier
```
wincent, line 5: Backticks please.

dxp/javascript_minification.md
```diff
@@ -0,0 +1,86 @@
...
+
+These three pieces conform the basic infrastructure available in DXP to minify JavaScript code.
```
wincent, line 7: "conform" → "form"

dxp/javascript_minification.md
```diff
@@ -0,0 +1,86 @@
...
+
+-   [JavaScriptMinifier](https://github.com/liferay/liferay-portal/blob/master/portal-impl/src/com/liferay/portal/minifier/JavaScriptMinifier.java): describes the interface that services to minify JavaScript must implement.
```
wincent, line 9: Backticks please.

dxp/javascript_minification.md
```diff
@@ -0,0 +1,86 @@
...
+
+-   [MinifierUtil](https://github.com/liferay/liferay-portal/blob/master/portal-impl/src/com/liferay/portal/minifier/MinifierUtil.java): provides a way to retrieve the `JavaScriptMinifier` with highest service ranking currently deployed to DXP. Note that you can also obtain a reference to a `JavaScriptMinifier` (or all of them) using standard OSGi mechanisms.
```
wincent, line 11: Backticks around `MinifierUtil` please.

> using standard OSGi mechanisms

What are those mechanisms? I have no idea. Could you turn that into a link somewhere informative?

dxp/javascript_minification.md
```diff
@@ -0,0 +1,86 @@
...
+
+-   [frontend-js-minifier](https://github.com/liferay/liferay-portal/tree/master/modules/apps/frontend-js/frontend-js-minifier): this project contains the two default minifiers that DXP ships out of the box (Google and Yahoo). See the two following sections for more details.
```
wincent, line 13: "out of the box" → "out-of-the-box"

dxp/javascript_minification.md
```diff
@@ -0,0 +1,86 @@
...
+
+### GoogleJavaScriptMinifier
```
wincent, line 15: Backticks.

dxp/javascript_minification.md
```diff
@@ -0,0 +1,86 @@
...
+
+This is the default DXP minifier (because it is deployed with a higher service ranking than the other ones available).
+
+It uses [Google Closure Compiler](https://github.com/google/closure-compiler) under the hood. Because this is an _internal_ library from Google (thus unsupported and subject to big breaking changes) updating it has usually a significant cost so it is not done very often.
+
+### YahooJavaScriptMinifier
```
wincent, line 21: Backticks.

dxp/javascript_minification.md
```diff
@@ -0,0 +1,86 @@
...
+
+-   [MinifierUtil](https://github.com/liferay/liferay-portal/blob/master/portal-impl/src/com/liferay/portal/minifier/MinifierUtil.java): provides a way to retrieve the `JavaScriptMinifier` with highest service ranking currently deployed to DXP. Note that you can also obtain a reference to a `JavaScriptMinifier` (or all of them) using standard OSGi mechanisms.
+
+-   [frontend-js-minifier](https://github.com/liferay/liferay-portal/tree/master/modules/apps/frontend-js/frontend-js-minifier): this project contains the two default minifiers that DXP ships out of the box (Google and Yahoo). See the two following sections for more details.
+
+### GoogleJavaScriptMinifier
+
+This is the default DXP minifier (because it is deployed with a higher service ranking than the other ones available).
+
+It uses [Google Closure Compiler](https://github.com/google/closure-compiler) under the hood. Because this is an _internal_ library from Google (thus unsupported and subject to big breaking changes) updating it has usually a significant cost so it is not done very often.
```
wincent, line 19: "has usually" → "usually has"

dxp/javascript_minification.md
```diff
@@ -0,0 +1,86 @@
...
+
+This uses [YUI Compressor](https://github.com/yui/yuicompressor) from Yahoo!, which is still active but unmaintained (except for some PRs that get merged from time to time). The last commit at the time of this writing (Sep 2020) is from May 2019.
+
+Unless someone has a very good reason to use it, this shouldn't be the choice for any production site but it is still shipped for compatibility reasons.
+
+## Components of DXP using the minifier
+
+Currently, the following components in DXP use the minifier services.
+
+### StripFilter
```
wincent, line 31: Backticks?

dxp/javascript_minification.md
```diff
@@ -0,0 +1,86 @@
...
+
+-   `/c/...`
+-   `/group/...`
+-   `/user/...`
+-   `/web/...`
+
+And it can be configured with:
+
+-   [portal.properties](https://github.com/liferay/liferay-portal/blob/master/portal-impl/src/portal.properties):
+    -   `minifier.inline.content.cache.enabled`: set to `true` to make the filter minify `<script>` tags content (by default `true`).
```
wincent, line 47: Comma after "default".

dxp/javascript_minification.md
```diff
@@ -0,0 +1,86 @@
...
+    -   `minifier.inline.content.cache.skip.javascript`: this defines a comma separated list of tokens to look for in the `<script>` content so that, if any is found, the minified response is not cached. By default, any `<script>` containing `getSessionId` or `encryptedUserId` is not cached.
+    -   `strip.ignore.paths`: list of paths to ignore (by default `/document_library/get_file`).
```
wincent, line 49: Comma after "default".

dxp/javascript_minification.md
```diff
@@ -0,0 +1,86 @@
...
+    -   `strip.mime.types` in `portal.properties`: defines the mime types of responses that should be processed by the filter (by default `text/html*` and `text/xml*`). Note that you can put a `*` at the end of the config values to catch all mime types starting with the string (not being exactly the string).
```
wincent, line 50: "mime" → "MIME" in two places

Comma after "default".

> not being exactly the string

This bit isn't clear to me. Does that mean that `*` matches 1 or more characters? (vs. 0 or more). If so, this might be more precise as:

> Note that you can put a `*` at the end of the config values to catch all mime types that have that value _as a prefix_ followed by additional characters (ie. `text/html*` does _not_ match `text/html` alone).

dxp/javascript_minification.md
```diff
@@ -0,0 +1,86 @@
...
+-   Request parameters:
+    -   `ensureContentLength`: set to `true` if you want to make sure the `Content-Length` header is returned from the server.
+    -   `strip`: set to `false` to disable the filter.
+    -   `p_p_lifecycle`: if it is `1` ([PortletRequest.ACTION_PHASE](https://docs.oracle.com/cd/E13155_01/wlp/docs103//javadoc/javax/portlet/PortletRequest.html#ACTION_PHASE)) or `2` ([PortletRequest.RESOURCE_PHASE](https://docs.oracle.com/cd/E13155_01/wlp/docs103//javadoc/javax/portlet/PortletRequest.html#RESOURCE_PHASE)) the filter is disabled.
```
wincent, line 54: Backticks around code things please.

dxp/javascript_minification.md
```diff
@@ -0,0 +1,86 @@
...
+
+If all the rules match and a `<script>` is minified, the filter simply invokes [MinifierUtil.minifyJavaScript()](https://github.com/liferay/liferay-portal/blob/master/portal-impl/src/com/liferay/portal/minifier/MinifierUtil.java#L45) to do its duty.
```
wincent, line 56: Exact line number here suggest you need "y" to make the link robust against future changes in `master`.

dxp/javascript_minification.md
```diff
@@ -0,0 +1,86 @@
...
+
+### AggregateFilter
```
wincent, line 58: Backticks please.

dxp/javascript_minification.md
```diff
@@ -0,0 +1,86 @@
...
+
+This servlet aggregates several `.js` or `.css` files and serves them as a single request to ease caching and reduce connection count. It also caches its responses.
+
+It honors the `work.dir.override.enabled` portal property so if it is set to `true` it diverts requests to development project folders when using fast deployments.
+
+The way to invalidate this servlet's cache is to check if any of the aggregated files has changed. If the portal property `combo.check.timestamp` is set to `false` (and by default it is in production systems) the check is not performed and the cache is never invalidated. Otherwise files are tested for changes every `combo.check.timestamp.interval` milliseconds (where `combo.check.timestamp.interval` is a configuration value of `portal.properties`).
+
+To finish with, each `.js` file aggregated by the servlet has its name tested for the `-min.js` or `.min.js` suffixes and, if they are NOT found, its content is minified using [MinifierUtil.minifyJavaScript()](https://github.com/liferay/liferay-portal/blob/master/portal-impl/src/com/liferay/portal/minifier/MinifierUtil.java#L45).
+
+### JSLoaderConfigServlet
```
wincent, line 72: Backticks please.

dxp/javascript_minification.md
```diff
@@ -0,0 +1,86 @@
...
+    -   `strip`: set to `false` to disable the filter.
+    -   `p_p_lifecycle`: if it is `1` ([PortletRequest.ACTION_PHASE](https://docs.oracle.com/cd/E13155_01/wlp/docs103//javadoc/javax/portlet/PortletRequest.html#ACTION_PHASE)) or `2` ([PortletRequest.RESOURCE_PHASE](https://docs.oracle.com/cd/E13155_01/wlp/docs103//javadoc/javax/portlet/PortletRequest.html#RESOURCE_PHASE)) the filter is disabled.
+
+If all the rules match and a `<script>` is minified, the filter simply invokes [MinifierUtil.minifyJavaScript()](https://github.com/liferay/liferay-portal/blob/master/portal-impl/src/com/liferay/portal/minifier/MinifierUtil.java#L45) to do its duty.
+
+### AggregateFilter
+
+This filter is already described in [Resource Serving in DXP](resource_serving.md#aggregate-filter).
+
+### ComboServlet
```
wincent, line 62: Backticks please.

dxp/javascript_minification.md
```diff
@@ -0,0 +1,86 @@
...
+
+This servlet services request to the `/js_loader_config` URL, which returns the [AMD Loader](https://github.com/liferay/liferay-amd-loader) configuration so that the browser side can access it.
+
+For example, it sets the following [loader properties](https://github.com/liferay/liferay-amd-loader/blob/master/src/loader/config.js): `explainResolutions`, `logLevel`, `resolvePath` and `waitTimeout`.
+
+It also controls the value of `Liferay.EXPLOSE_GLOBAL` which causes `Liferay.Loader.define()` method to be published under `window.define` (this is for compatiblity reasons, but its use is strongly discouraged as it can cause issues with third party libraries).
+
+Even though it is small, this servlet minifies its response with [MinifierUtil.minifyJavaScript()](https://github.com/liferay/liferay-portal/blob/master/portal-impl/src/com/liferay/portal/minifier/MinifierUtil.java#L45) and also caches it for efficiency.
+
+### JSBundleConfigServlet
```
wincent, line 82: Backticks please.

dxp/javascript_minification.md
```diff
@@ -0,0 +1,86 @@
...
+
+### ComboServlet
+
+This servlet aggregates several `.js` or `.css` files and serves them as a single request to ease caching and reduce connection count. It also caches its responses.
+
+It honors the `work.dir.override.enabled` portal property so if it is set to `true` it diverts requests to development project folders when using fast deployments.
+
+The way to invalidate this servlet's cache is to check if any of the aggregated files has changed. If the portal property `combo.check.timestamp` is set to `false` (and by default it is in production systems) the check is not performed and the cache is never invalidated. Otherwise files are tested for changes every `combo.check.timestamp.interval` milliseconds (where `combo.check.timestamp.interval` is a configuration value of `portal.properties`).
+
+To finish with, each `.js` file aggregated by the servlet has its name tested for the `-min.js` or `.min.js` suffixes and, if they are NOT found, its content is minified using [MinifierUtil.minifyJavaScript()](https://github.com/liferay/liferay-portal/blob/master/portal-impl/src/com/liferay/portal/minifier/MinifierUtil.java#L45).
```
wincent, line 70: "To finish with" → "Finally"

Also another line-specific link here that could be robustified by pointing it at a concrete blob.

dxp/javascript_minification.md
```diff
@@ -0,0 +1,86 @@
...
+
+It also controls the value of `Liferay.EXPLOSE_GLOBAL` which causes `Liferay.Loader.define()` method to be published under `window.define` (this is for compatiblity reasons, but its use is strongly discouraged as it can cause issues with third party libraries).
+
+Even though it is small, this servlet minifies its response with [MinifierUtil.minifyJavaScript()](https://github.com/liferay/liferay-portal/blob/master/portal-impl/src/com/liferay/portal/minifier/MinifierUtil.java#L45) and also caches it for efficiency.
+
+### JSBundleConfigServlet
+
+This servlet services request to the `/js_bundle_config` URL, which is the configurator of the legacy `AUI` loader.
+
+This servlet does not cache its response but minifies it leveraging [MinifierUtil.minifyJavaScript()](https://github.com/liferay/liferay-portal/blob/master/portal-impl/src/com/liferay/portal/minifier/MinifierUtil.java#L45).
```
wincent, line 86: Another link that needs "y" treatment.

wincent (commented):
Looks great!

dxp/javascript_minification.md
izaera: :thinking: I'm pointing things to `master` because they are fairly stable and, in case of small changes, you see the newest copy and it can enrich the documentation. 

This has the drawback that the link can also be broken, but it is very probable that if the links break because the files have been (re)moved the documentation becomes obsolete, so I decided in the end to use `master` links as the best trade-off...

izaera: Another option is not pointing to the line but to the file (I used that technique in some other places...)

wincent: > 🤔 I'm pointing things to `master` because they are fairly stable and, in case of small changes, you see the newest copy and it can enrich the documentation.

This is a bad idea. It risks destroying information, or at least makes it hard to reconstruct:

-   If `master` changes, I can land in the file and see that the line I'm at obviously isn't the one you were talking about.
-   I then have to perform a difficult search to find out what master looked like at the time you published this document, which means looked at the date the line entered the documentation repo, and then finding the corresponding hash on master.
-   Worse still, if we rename or move the files (which has happened multiple times in recent years even for files that don't themselves get touched, as we move massive parts of the directory tree up or down the hierarchy) then finding the document you linked to is even more cumbersome.

Contrast that with the opposite, which I am suggesting:

-   I land on "stale" documentation. I see a URL like https://github.com/liferay/liferay-portal/blob/ab15fb32bd58e4f218c8270be54ba8a835e253b7/portal-impl/src/com/liferay/portal/minifier/MinifierUtil.java#L45 with a hash, so I know I may not looking at current `master`.
-   Based on the header metadata, I decide whether I want to look to see for the current state:

    Screenshot 2020-09-11 -154310-87BzyZe9@2x
-   If I do, I double-click the SHA1 in the URL bar, type `master`, hit Enter.
-   If stuff has moved, I can hunt around for it.

I previously [managed to convince](https://github.com/liferay/liferay-frontend-guidelines/pull/29#discussion_r293012635) @julien and @jbalsas to do this (although I can't find an example right now, I've seen him giving people this same advice). I probably worded it better back then:

> I suggest hitting "y" on GitHub when you're grabbing links like:
>
> That turns it into an "evergreen" URL that will never break unless we rewrite history (which we won't):
>
> It's all too easy for these links to become dead and broken at the next rename, and that's worse than looking at something slightly stale, I think.

Evergreen links are better, and one of the great things about an immutable version control system is you actually can post links that are immune to bit-rot.

izaera: The only problem with that reasoning is that some people may not notice that they are on a given commit, instead of master's HEAD and think that code is the latest.

But it is true that it looks like a better scenario, still, so I'll change live links to fixed ones.

The thing is, if we decide to go that way, we should probably use fixed links for everything and don't mix them, to avoid fooling readers.

If, for some reason, we want to use a live link in some place, we should state in that place that the link is live and may change without the documentation being updated.

Does that sounds like a good strategy for the whole repo? :thinking:

jbalsas: > Does that sounds like a good strategy for the whole repo? 🤔

I've been trying to paste fixed links for a while now. They do feel safer in the sense that they reflect the state of the code at the time of writing, which makes the article coherent at the very least.

While the code might be stale, so does the article, so this sounds to me like a more robust strategy.

dxp/javascript_minification.md
jbalsas: How do the `portal-impl` classes access these `module` minifiers? I might be wrong here, but I assume we didn't remove the minifiers in `portal-impl`... could we make some notes about that, just to clarify what's being used and what isn't?

dxp/javascript_minification.md
wincent: My general policy:

- If I am linking to a specific line, _always_ use a fixed link.
- If I am linking to a specific file, use my judgment:
  - If it is possible that the file might be renamed (eg. any source code file), I prefer fixed links.
  - If it is extremely unlikely that a file will be renamed (eg. a `package.json` or a top-level `README.md`), I don't use a fixed link.

I prefer simple rules (not of the form "if A then B else C unless D") because they are easily explained, easily understood, easily enforced etc so it pains me to write that out... you can kind of squint at this one and just reduce it to "favor using fixed links".

forced-smile

dxp/javascript_minification.md
izaera: > How do the portal-impl classes access these module minifiers

Through `MinifierUtil`, because it is in `portal-impl`.

jbalsas: I mean... we have a static `util` in `portal-impl` accessing OSGi Services?

jbalsas: I see we don't have them in `portal-impl` anymore, though, so there's no duplication which was my main source of confusion.

izaera: I added some text explaining it.

izaera: > I mean... we have a static util in portal-impl accessing OSGi Services?

Yes, that was the usual way of OSGifying things in the past without breaking the API. That's the reason why OSGi `Registry` was created: to bridge outside-of-OSGi code and OSGi world.

support/checkLinks.js
```diff
@@ -88,6 +88,14 @@ async function checkInternal(link, files) {
 }
 
 async function checkLocal(link, files) {
+	const i = link.indexOf('#');
+	let anchor;
+
+	if (i != -1) {
```
wincent, line 94: Prefer `!==` over `!=`.

In this case though, maybe we could do:

```js
	const [base, fragment] = link.split('#');

	for (const file of files) {
		let target;

		if (path.isAbsolute(base)) {
			// Resolve relative to repo root.
			target = path.join(__dirname, '..', link);
		} else {
			// Resolve relative to current file's directory.
			target = path.join(path.dirname(file), base);
		}

		try {
			await accessAsync(target);
		} catch (error) {
			report(file, `No file/directory found for local target: ${target}`);
		}

		if (fragment) {
			await checkInternal(fragment, new Set([target]));
		}
```

wincent (approved):
> refactor: make previous fix cleaner

😍

PR #209, opened 2020-09-16
docs: add the Build SCSS to the DXP Build Process document

matuzalemsteles:


dxp/build_process.md
```diff
@@ -60,7 +60,7 @@ The CLI can also run [webpack](https://webpack.js.org/) as part of the process i
 
 #### 6. Liferay npm bundler
 
-Running [liferay-npm-bundler](https://github.com/liferay/liferay-js-toolkit/tree/master/packages/liferay-npm-bundler) allows your project's dependencies and source code to be converted to [AMD}(https://en.wikipedia.org/wiki/Asynchronous_module_definition) so that they can be deployed as an OSGi package later. So, Liferay Portal will be able to load the files of your module and its dependencies when requested.
+Running [liferay-npm-bundler](https://github.com/liferay/liferay-js-toolkit/tree/master/packages/liferay-npm-bundler) allows your project's dependencies and source code to be converted to [AMD](https://en.wikipedia.org/wiki/Asynchronous_module_definition) so that they can be deployed as an OSGi package later. So, Liferay Portal will be able to load the files of your module and its dependencies when requested.
```
wincent, line 63: That `}` ... 😱

noooooo

dxp/build_process.md
```diff
@@ -72,6 +72,65 @@ The `liferay-npm-bridge-generator` is only executed when there is a `.npmbridger
 
 > To learn more about how to use this tool read ["How to use liferay npm bridge generator" in the wike of the liferay-js-toolkit repository](https://github.com/liferay/liferay-js-toolkit/wiki/How-to-use-liferay-npm-bridge-generator).
 
+### Build SCSS
+
+There are two ways that DXP handles building `.scss` files:
+
+-   [CSS Builder Gradle Plugin](https://github.com/liferay/liferay-portal/blob/master/modules/sdk/gradle-plugins-css-builder/README.markdown)
+-   Liferay npm bundler with [sass-loader](https://github.com/liferay/liferay-js-toolkit/tree/master/packages/liferay-npm-bundler-loader-sass-loader) and [css-loader](https://github.com/liferay/liferay-js-toolkit/tree/master/packages/liferay-npm-bundler-loader-css-loader) via JavaScript
```
wincent, line 80: I think you should drop the reference to the sass-loader, because we don't use it, and AFAIK, don't want to use it:

```
$ yarn why liferay-npm-bundler-loader-sass-loader
yarn why v1.22.5
[1/4] 🤔  Why do we have the module "liferay-npm-bundler-loader-sass-loader"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
error We couldn't find a match!
```

dxp/build_process.md
```diff
@@ -72,6 +72,65 @@ The `liferay-npm-bridge-generator` is only executed when there is a `.npmbridger
...
+
+Gradle plugin is executed together with the `gradlew deploy` build pipeline and expects the `.scss` files in the module's `src/main/resources` folder. Compiled by the [`CSSBuilder`](https://github.com/liferay/liferay-portal/blob/master/modules/util/css-builder/src/main/java/com/liferay/css/builder/CSSBuilder.java) which is responsible for the build and [`CSSRTLConverter`](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-css/frontend-css-rtl-servlet/src/main/java/com/liferay/frontend/css/rtl/servlet/internal/converter/CSSRTLConverter.java) to create the [RTL](https://en.wikipedia.org/wiki/Right-to-left) files for the final css. The module portlet defines the path to the final CSS file that adds the CSS path to the HTML output of the request.
```
wincent, line 82: - "The Gradle plugin"
- Drop the "the" before `CSSBuilder`, unless you're going to say something like "the `CSSBuilder` class".
- "css" → "CSS".

dxp/build_process.md
```diff
@@ -72,6 +72,65 @@ The `liferay-npm-bridge-generator` is only executed when there is a `.npmbridger
...
+
+Unlike the Gradle plugin, loaders are executed in the Liferay npm bundler, which consequently is part of the build flow of `liferay-npm-scripts`, the difference is that you can import a `.css` or `.scss` file in JavaScript file .
```
jbalsas, line 84: This made me realize we don't do the `_rtl` conversion for these files... 🤔

jbalsas:
LGTM!

I'll leave @wincent comment on the grammar 🤓

dxp/build_process.md
```diff
@@ -72,6 +72,65 @@ The `liferay-npm-bridge-generator` is only executed when there is a `.npmbridger
...
+
+```js
+// When `sass-loader` is configured
+import './button.scss';
+
+// When `css-loader` is configured
+import './button.css';
+```
+
+The output of the loaders creates a JavaScript file with the name of the CSS file including the suffix `scss` or `css` that takes care of loading the final CSS file, similar to this:
```
wincent, line 94: Here again, I'd say don't mention `sass-loader` because we don't want people to use it.

If you do a `git grep 'import.+\.s?css'` in the liferay-portal repo, you will find people importing both CSS and SCSS files, but the ".scss" files aren't being processed by the sass-loader (as `yarn why` above shows) — they are getting handled by the css-loader [as described here](https://github.com/liferay/liferay-npm-tools/issues/239):

> For now, the gradle build will compile from .scss to .css and the bundler will be just responsible of including the .css file with `css-loader`.

As for the places doing `import 'something.css'`, they are also being handled by the `css-loader` due to [this default bundler config](https://github.com/liferay/liferay-npm-tools/blob/c52308eeb719222e63e32f1abea88426011581fa/packages/liferay-npm-bundler-preset-liferay-dev/config.json#L270-L287) which applies to most places in liferay-portal:

```
	"rules": [
		{
			"test": "\\.css$",
			"use": ["css-loader"]
		},
		{
			"test": "\\.scss$",
			"exclude": "node_modules",
			"use": [
				{
					"loader": "css-loader",
					"options": {
						"extension": ".css"
					}
				}
			]
		}
	],
```

There are two other places in liferay-portal that configure the `css-loader` (see `git grep css-loader -- '**/.npmbundlerrc'`). I don't know whether that config is actually needed or not; I haven't checked to see if they are using the above preset, so I would just ignore them for the purposes of this analysis.

@izaera will correct me if I'm wrong on any of this.

wincent (changes requested):

dxp/build_process.md
jbalsas: Thanks for this, @wincent, this made me realize that what I say in https://github.com/liferay/liferay-frontend-guidelines/pull/209#discussion_r490027818 is wrong :)

We still rely on `buildCSS` to build both the `css` and `css_rtl` files for a given `.scss` file. That is, all `scss` files are processed by our gradle plugin. 

We only use `css-loader` because we can expect all files to be there ready in the first place 🙏

dxp/build_process.md
jbalsas: See https://github.com/liferay/liferay-frontend-guidelines/pull/209#discussion_r490036526 for some clarification... it's important to note that we go through `CSSBuilder` in any case since that's got us covered and then just use the `css-loader` on an already-processed file.

Phew! 😂

dxp/build_process.md
izaera: there there

dxp/build_process.md
izaera: > don't want to use it

:cry: That hurt...

dxp/build_process.md
izaera: Yes. The CSS processing is done by gradle and then the bundler takes on from there. We created the SASS loader as a PoC but never integrated it, though we left it in the toolkit because it makes sense on its own.

dxp/build_process.md
wincent: tough-love

dxp/build_process.md
matuzalemsteles: 😂😂😂

dxp/build_process.md
matuzalemsteles: 😱 oh sorry, what a pity! I'm sorry @izaera haha

dxp/build_process.md
matuzalemsteles: Thanks for the info! I will update.

dxp/build_process.md
matuzalemsteles: haha

matuzalemsteles:
@wincent I updated the document, I ended up changing some things with the information that we are still using the Gradle plugin independent of the loader, so I wrote from another perspective that we have two ways to load the final files.

wincent (approved):
This looks good @matuzalemsteles. I am just going to apply some tiny English tweaks on top and merge it (hopefully won't introduce a syntax error this time).

wincent:
CI likes it, so I like it.

PR #211, opened 2020-09-18
WIP: docs: importable-css in dxp

eduardoallegrini:
WIP

What CSS can I import into my module and theme and how to do it?

### DIFF
--- a/dxp/what-css-can-i-import-into-my-module-and-theme.md
+++ b/dxp/what-css-can-i-import-into-my-module-and-theme.md
@@ -2,6 +2,8 @@
 
 CSS libraries that are available in DXP and how to import in your module or theme
 
+## atlas
+
 ### What does it provide?
 Table of variables, mixins, functions
 
@@ -25,7 +27,7 @@ buildCSS {
 ### How to use it?
 Example of code
 
-## atlas
+## atlas-variables
 
 ## clay
 
--- a/dxp/what-css-can-i-import-into-my-module-and-theme.md
+++ b/dxp/what-css-can-i-import-into-my-module-and-theme.md
@@ -1,0 +1,57 @@
+# What CSS can I import into my module and theme?
+
+CSS libraries that are available in DXP and how to import in your module or theme
+
+### What does it provide?
+Table of variables, mixins, functions
+
+### How to import it?
+
+In your main SCSS file
+```scss
+@import 'atlas';
+```
+
+In your build.gradle
+```js
+buildCSS {
+	imports = [
+		new File(npmInstall.nodeModulesDir, "@clayui/css/src/scss")
+	]
+}
+```
+
+
+### How to use it?
+Example of code
+
+## atlas
+
+## clay
+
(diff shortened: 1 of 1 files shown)


dxp/what-css-can-i-import-into-my-module-and-theme.md
```diff
@@ -0,0 +1,57 @@
...
+CSS libraries that are available in DXP and how to import in your module or theme
+
+### What does it provide?
+Table of variables, mixins, functions
+
+### How to import it?
+
+In your main SCSS file
+```scss
+@import 'atlas';
```
wincent, line 12: I thought we didn't want to encourage this (importing all of `atlas`); I thought we wanted to encourage importing `atlas-variables` instead.

We will be [adding a lint for this](https://github.com/liferay/liferay-npm-tools/issues/479), but obviously it won't run over the documentation.

eduardoallegrini: The goal of this PR is to explain (and understand) why and how to import these CSS libraries. It will be easier to encourage the importing of `atlas-variables` knowing what exactly does `@import 'atlas'` 🤷‍♂️

wincent: It's dangerous to leave erroneous code lying around, even as a counterexample. If someone skims it without really understanding it, you've just made the problem worse.

eduardoallegrini: I think we shouldn't consider it erroneous, sometimes we'll need `atlas` instead of `atlas-variables` and if someone wants to try it or use for different projects outside DXP it could be very useful if we explain how and when to use it.

In my opinion, we should provide the information about options like `atlas`, `base` or `base-variables` that are not used in DXP, but do exist in Clay-CSS and to be sure we can add a big red alert banner that says "don't use it unless you understand what you're doing" and the lint you are working on. 😆

jbalsas: The key at this point was not to document what can be used but how things work. That would help us inform what to use in the future.

For example, what I think would be interesting to document is:
- That each module does an independent sass compilation unit. This is partly covered in https://github.com/liferay/liferay-frontend-guidelines/blob/master/dxp/build_process.md#build-scss but could be expanded to clarify the implications.
- How is Clay brought into DXP via Themes and how it is injected
- How is the CSS for tags or other common components compiled and used in the page
- What are the Portlet CSS headers, how are they used and how they insert CSS in the page

Having a clear overview of this may allow us to propose architecture changes based on sound knowledge and understanding of the existing platform.

> I think we shouldn't consider it erroneous, sometimes we'll need atlas instead of atlas-variables and if someone wants to try it or use for different projects outside DXP it could be very useful if we explain how and when to use it.

While that's true, keep in mind you called this `importable-css in DXP`, so I think "projects outside DXP" would be out of the scope of this document.

PR #213, opened 2020-09-21
docs: How does Clay arrive to DXP

eduardoallegrini:
This article explains how Clay arrives to DXP

PR #215, opened 2020-09-23
docs: Clay CSS 3.x introduction

pat270:
Requested by @jbalsas, not sure what directory to place this in. I feel like this information is/will be duplicated in a lot of other documents (#213).

/cc @marcoscv-work @eduardoallegrini

### DIFF
--- a/dxp/clay-css-3-intro.md
+++ b/dxp/clay-css-3-intro.md
@@ -1,0 +1,194 @@
+# Introduction to @clayui/css (3.x)
+
+Clay CSS is the web implementation of <a href="https://liferay.design/lexicon/get-started/" rel="noreferrer noopener" target="_blank">Liferay’s Experience Language</a>. This is a system for building applications in and outside of Liferay, designed to be fluid and extensible, as well as provide a consistent and documented API.
+
+Clay CSS is based on <a href="https://getbootstrap.com/docs/4.4/getting-started/introduction/" rel="noreferrer noopener" target="_blank">Bootstrap 4.4.1</a>. You will find Bootstrap's and Clay CSS' API to be very similar because Clay CSS was created to fill the gaps between Bootstrap's and Liferay's UI as well as providing familiarity to new devs joining the Liferay ecosystem.
+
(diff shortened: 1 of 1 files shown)


wincent:
> I feel like this information is/will be duplicated in a lot of other documents (#213).

I don't know how urgent @jbalsas considers this to be, but maybe we can pause one of these two PRs so that we can get one in first, and then make sure that the second one links out to the other one where appropriate instead of duplicating the same information in a different format.

matuzalemsteles:
This is a type of documentation that would be great to be in the [Clay repository](https://github.com/liferay/clay/tree/master/docs) but we can just create a document that links to this one.

pat270:
I don't mind pausing this one since #213 was first or adding something like this to the Clay repo. We can see how it plays out.

@matuzalemsteles I added a wiki https://github.com/liferay/clay/wiki/Liferay-Portal-@clayui-css-Versions for @jonmak08 yesterday, would it be better in docs?

matuzalemsteles:
> @matuzalemsteles I added a wiki https://github.com/liferay/clay/wiki/Liferay-Portal-@clayui-css-Versions for @jonmak08 yesterday, would it be better in docs?

I see no problem with this being on the Wiki if it is the case of easy access for people but I have the impression that being in docs I seem easier to track.

wincent:
> would it be better in docs?

The wiki is easier to update but less discoverable (because I think people are generally "trained" to seek the docs first), but you could link to your wiki article from the docs and have the best of both worlds.

eduardoallegrini:
🤔 I took a look at the documentation and I think we can have both pages without duplicating the info. I'm trying to avoid repeating what you already wrote in this PR and leaving the space to link them once ready.

pat270:
@eduardoallegrini I think it's ok to repeat. Personally, I like having all the info I need in one place rather than having to click around. I get easily lost.

jonmak08:
Just adding my two cents... Doing what @wincent mentioned might seem to be the best way to have it versioned and updated, while also having the wiki information linked.  Does that approach not work?

jbalsas:
I think we can simply merge this. We can always move it elsewhere or link it in the future.

My only doubt is wether `/css` is the folder for it... thoughts?

pat270:
I can move this to the DXP folder. I don't really have an opinion on where it should live.

PR #217, opened 2020-09-24
docs: DXP minimizing bleeding themes

pat270:


dxp/how_to_minimize_bleeding_themes.md
```diff
@@ -0,0 +1,360 @@
+# How to Minimize Bleeding Themes
+
+> "Currently upgrading from DXP 7.0 to DXP 7.2 and they have a leaking theme. What I mean is that some styling they apply to the site also has an effect within the control panel."
```
wincent, line 3: You don't need the double quotes here... `>` already is a quote, right?

dxp/how_to_minimize_bleeding_themes.md
```diff
@@ -0,0 +1,360 @@
...
+
+This is a common problem due to the global nature of CSS and Liferay's complexity. This document will outline some methods to minimize bleeding CSS.
+
+## Working With Clay CSS
+
+Liferay Portal admin controls use Clay CSS components extensively and aren't namespaced so modifying Clay CSS components directly will cause unintended issues. The way we can work around this is by namespacing our CSS selectors to only target the components we want to change.
+
+We shouldn't style HTML elements directly. For example:
+
+```css
```
wincent, line 13: For these things that are actually _counter_-examples...

counter-example

... could we flag them explicitly like this (or similar)?

---

**⚠️ Counterexample:**

```css
a {
    color: orange;
}

/* etc ... */
```

pat270:
I made the changes, thanks @wincent !

wincent (approved):
Well, seeing as nobody else has any complaints, let's merge it!

jbalsas:
iobject

(I'm really just sleep deprived :D )

jbalsas:
I think this is good material for people getting hit by this in `7.3`. For `7.4`, I wanted to explore ShadowDOM or other approaches to shield admin from content and avoid this altogether. You should be able to find some initial task explorations in our Roadmap!

wincent:
> (I'm really just sleep deprived :D )

I figured you were going to come along and say it would be nice to talk about Shadow DOM or some other way of isolating styles. Or that this should have been in the `css/` directory. Or just something like this:

a-mess

jbalsas:
> I figured you were going to come along and say it would be nice to talk about Shadow DOM or some other way of isolating styles. Or that this should have been in the css/ directory. Or just something like this:

I actually did, didn't I? 😂

PR #223, opened 2020-09-30
docs: Senna.js integration in DXP 

julien:
@matuzalemsteles and I worked on this to document Senna.js integration in DXP.
It probably still need some changes/improvements, but we think this is good for a first review


dxp/sennajs_integration.md
```diff
@@ -0,0 +1,117 @@
...
+
+This article describes how [Senna.js](https://github.com/liferay/senna.js) is integrated in DXP and how it works under the hood.
+
+### Introduction
+
+The infrastructure around Senna.js is located in the [`frontend-js-spa-web`](https://github.com/liferay/liferay-portal/tree/2d4120b7fdb8cf03e7505ebdaacee3bdceec3391/modules/apps/frontend-js/frontend-js-spa-web) module.
+
+In this module, a Senna.js [`App`](https://github.com/liferay/senna.js/blob/059022b737de0f0108b7ef3f06f14d7ade3baa94/src/app/App.js) instance is created, with two routes which are handled by a specific [`Screen`](https://github.com/liferay/senna.js/blob/059022b737de0f0108b7ef3f06f14d7ade3baa94/src/screen/Screen.js) instance.
+
+-   The [`ActionURLScreen`](https://github.com/liferay/liferay-portal/blob/2d4120b7fdb8cf03e7505ebdaacee3bdceec3391/modules/apps/frontend-js/frontend-js-spa-web/src/main/resources/META-INF/resources/liferay/screen/ActionURLScreen.es.js) class is used for all requests made to "action" URLs, which are used to post forms. It makes sure that the http method used is `POST` and checks that the `p_p_lifecycle` URL parameter has the correct value when submitting a form.
```
wincent, line 11: > has the correct value

Should we mention what that value is? Off the top of my head, I don't know.

dxp/sennajs_integration.md
```diff
@@ -0,0 +1,117 @@
...
+
+Both of these classes inherit from the [`EventScreen`](https://github.com/liferay/liferay-portal/blob/2d4120b7fdb8cf03e7505ebdaacee3bdceec3391/modules/apps/frontend-js/frontend-js-spa-web/src/main/resources/META-INF/resources/liferay/screen/EventScreen.es.js) class which adds specific behaviour on top of Senna.js's [`Screen`](https://github.com/liferay/senna.js/blob/059022b737de0f0108b7ef3f06f14d7ade3baa94/src/screen/Screen.js) class such as dispatching events during the screen's lifecycle (loading, activated, disposed) or managing the cache.
+
+The Senna.js `App` instance is exposed through the global variable [`Liferay.SPA.app`](https://github.com/liferay/liferay-portal/blob/2d4120b7fdb8cf03e7505ebdaacee3bdceec3391/modules/apps/frontend-js/frontend-js-spa-web/src/main/resources/META-INF/resources/liferay/init.es.js#L129) although that might change in the future
```
wincent, line 17: You're missing a full-stop at the end of this sentence.

What makes us think that this might change in the future? (I know we're trying to get rid of globals, but this one seems particularly unlikely to go away soon.)

dxp/sennajs_integration.md
```diff
@@ -0,0 +1,117 @@
...
+
+The `frontend-js-spa-web` modules also contains templates to [configure](https://github.com/liferay/liferay-portal/blob/2d4120b7fdb8cf03e7505ebdaacee3bdceec3391/modules/apps/frontend-js/frontend-js-spa-web/src/main/resources/META-INF/resources/config.tmpl) and [initialize](https://github.com/liferay/liferay-portal/blob/2d4120b7fdb8cf03e7505ebdaacee3bdceec3391/modules/apps/frontend-js/frontend-js-spa-web/src/main/resources/META-INF/resources/init.tmpl) the senna.js `App`.
```
wincent, line 19: "modules" → "module"

dxp/sennajs_integration.md
```diff
@@ -0,0 +1,117 @@
...
+
+Everything is included with the [SPATopHeadJSPDynamicInclude](https://github.com/liferay/liferay-portal/blob/2d4120b7fdb8cf03e7505ebdaacee3bdceec3391/modules/apps/frontend-js/frontend-js-spa-web/src/main/java/com/liferay/frontend/js/spa/web/internal/servlet/taglib/SPATopHeadJSPDynamicInclude.java#L134-L143) if DXP is configured to use SPA, which it is by default (except in [Internet Explorer](https://github.com/liferay/liferay-portal/blob/454d7f6a439dc668a0ce0529cff2cae87ee9a0d5/modules/apps/frontend-js/frontend-js-spa-web/src/main/java/com/liferay/frontend/js/spa/web/internal/configuration/SPAConfiguration.java#L38-L50)).
```
wincent, line 21: Backticks around `SPATopHeadJSPDynamicInclude`.

dxp/sennajs_integration.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Block list
+
+It's also possible to tell the Senna.js `App` instance which portlets it needs to ignore, in which case it will not be used for navigation or form submission.
+
+Any portlet that is set to not use "Single Application Mode", or any portlet that is not deployed (and not ready) is automatically added to this list.
+
+## Cache
+
+The Senna.js `App` will use caching, if the `Liferay.SPA.cacheExpirationTime` variable's value is greater than `-1`.
```
wincent, line 44: Really? So if I set it to `-0.9` it will still cache?

dxp/sennajs_integration.md
```diff
@@ -0,0 +1,117 @@
...
+By default DXP sets this value to `-1` which means that the cache is disabled, if you want Senna.js cache to never be cleared you can set this value to `0`
+or specify the value (in minutes) after when Senna.js cache should be cleared.
+
+The cache expiration time can also be configured by navigating to **Control Panel** > **System Settings** > **Infrastructure** > **Frontend SPA Infrastructure** and changing the value specified in the **Cache Expiration Time** field.
+
+With caching enabled, once a resource is loaded, it's content is cached into memory and is retrieved later on without any additional request.
```
wincent, line 50: "into memory" → "in memory"

dxp/sennajs_integration.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Excluded paths
+
+The `frontend-js-spa-web` module also introduces a property called [`spa.exluced.paths`](https://github.com/liferay/liferay-portal/blob/454d7f6a439dc668a0ce0529cff2cae87ee9a0d5/modules/apps/frontend-js/frontend-js-spa-web/src/main/resources/spa.properties) which, as it's name indicates, is a list of paths that should be ignored by the `Senna.js` App instance.
```
wincent, line 54: "it's" → "its"

dxp/sennajs_integration.md
```diff
@@ -0,0 +1,117 @@
...
+
+These paths, are visible in the global `Liferay.SPA.excludedPaths` variable.
+
+## URLs
+
+By default Senna.js in DXP will contain the required parameters needed by portlets, specifically `p_p_id`, `p_p_lifecycle`, `p_p_state` and `p_p_auth`
```
wincent, line 60: Need a full-stop at the end of this sentence.

dxp/sennajs_integration.md
```diff
@@ -0,0 +1,117 @@
...
+
+## How Senna.js works
+
+![Simplified overview of Senna.js](../images/sennajs-overview.png)
+
+Senna.js takes into account concepts for dealing with SPA on sites that are not built purely on a framework/library, such as React, Vue or Angular.
```
wincent, line 66: This sentence isn't all that clear; I think I'd just remove it.

dxp/sennajs_integration.md
```diff
@@ -0,0 +1,117 @@
...
+
+Senna.js, introduces the following components:
```
wincent, line 68: Can we call them "concepts" instead of "components"? (Otherwise they sound a bit like "UI components")

dxp/sennajs_integration.md
```diff
@@ -0,0 +1,117 @@
...
+
+-   **Routes** Determine which path will be controlled by Senna.js and which Screen will handle the content.
+-   **Surface** A DOM element with an `id` attribute, which is where the content will be rendered.
+-   **Screen** A special type of route handler to handle the navigation request.
+
+To simulate a single page application, Senna.js [tracks all links](https://github.com/liferay/senna.js/blob/master/src/app/App.js#L104) within the Surface, when a link is triggered, Senna.js controls the event by preventing the default behavior to prevent navigation from being made by the Browser and then triggers the navigation process.
```
wincent, line 74: At this is a link to a line number that may change, can we make it permanent by including a hash? (ie. hit "y" in the GitHub UI.)

Also, we have a run on sentence here, so put a full-stop after "within the Surface".

"Browser" → "browser" because it is not a proper noun or even a special concept in this document.

dxp/sennajs_integration.md
```diff
@@ -0,0 +1,117 @@
...
+2. Invoke the Screen that handles this route
+3. Prepares the surface and sequentially calls the life cycle of a Screen
+    - load
+    - getTitle
+    - beforeUpdateHistoryPath
+    - beforeUpdateHistoryState
+    - getSurfaceContent
+    - evaluateStyles
+    - flip
+    - evaluateScripts
```
wincent, line 92: These should all have backticks.

dxp/sennajs_integration.md
```diff
@@ -0,0 +1,117 @@
...
+4. Synchronizes the scroll position if necessary
+5. Finalize navigation
+    - Activate screen
+    - Caches the Screen in memory
+
+The Screen can control the entire loading and extraction cycle of content resources to allocate in the DOM. [`HtmlScreen`](https://github.com/liferay/senna.js/blob/master/src/screen/HtmlScreen.js) is responsible for a good part of the manipulation and extraction of information from the content of the request and extends the [`RequestScreen`](https://github.com/liferay/senna.js/blob/master/src/screen/RequestScreen.js) that takes care of setting up the request of the requested path.
```
wincent, line 98: No "the" before `RequestScreen`.

dxp/sennajs_integration.md
```diff
@@ -0,0 +1,117 @@
...
+
+```
+HtmlScreen -> RequestScreen -> Screen
+```
+
+When the load happens all the returned content is allocated in a virtual document, Screen does not add directly to the DOM to not cause problems and for a better experience during the transition, the transition to the page is done only when everything is ready, style, scripts, and elements.
```
wincent, line 104: Instead of:

> when everything is ready, style, scripts, and elements.

Do:

> when everything is ready (styles, scripts, and elements).

dxp/sennajs_integration.md
```diff
@@ -0,0 +1,117 @@
...
+
+### Resource tracking
+
+![](../images/sennajs-flip.png)
+
+Resource tracking happens when navigation is being made, this includes the `style` and `script` tags, Screen looks for those elements that are marked with the `data-senna-track` attribute within the content that is in a virtual document or in the DOM, there are two types of tracking for these attributes.
```
wincent, line 110: Another run-on sentence.

> Resource tracking of `style` and `script` tags happens during navigation. `Screen` looks for those elements that are marked with the `data-senna-track` attribute within the content that is in a virtual document or in the DOM.
>
> There are two types of tracking for these attributes:

dxp/sennajs_integration.md
```diff
@@ -0,0 +1,117 @@
...
+
+-   **Permanent** These are resources that persist during navigation regardless of the route
+-   **Temporary** These are resources that are removed in the next route navigation
+
+> Any resource that has this attribute declared and is within the document's `body` will be moved to the document's `head`.
+
+The resources need to be loaded before proceeding with the next stages of navigation but that does not mean that Senna.js will wait for JavaScript and CSS to be evaluated, this only counts the loading of the source code and it may be normal to see the transition to a blank canvas because that page is rendered using React for example.
```
wincent, line 117: Full-stop after "evaluated".

wincent (commented):
Some little nits inline, but for me (as a non-Senna-expert) this is a very useful guide.

dxp/sennajs_integration.md
julien: Yes that's actually a good idea, the value should be `0` - or at least that's what I understand when I read [this](https://github.com/liferay/liferay-portal/blob/b619978445e3be09526803b27447b5ddab95ca6f/modules/apps/frontend-js/frontend-js-spa-web/src/main/resources/META-INF/resources/liferay/screen/ActionURLScreen.es.js#L53-L55)

dxp/sennajs_integration.md
julien: You read my mind, I actually thought that we might be able to get rid of this global - without even searching for usages :see_no_evil: 
As you said; this one will probably be hard to get rid of, so I might as well remove that.

dxp/sennajs_integration.md
julien: It [looks](https://github.com/liferay/liferay-portal/blob/021b09a2df5e71140dd6061497000db108f09048/modules/apps/frontend-js/frontend-js-spa-web/src/main/resources/META-INF/resources/liferay/app/App.es.js#L136-L138) like it will

dxp/sennajs_integration.md
julien: I always get that one wrong - it's a shame

dxp/sennajs_integration.md
julien: Yes of course, I just didn't come up with a better name - but it's true that "components" is confusing in this case.

dxp/sennajs_integration.md
wincent: Although [here](https://github.com/liferay/liferay-portal/blob/021b09a2df5e71140dd6061497000db108f09048/modules/apps/frontend-js/frontend-js-spa-web/src/main/resources/META-INF/resources/liferay/app/App.es.js#L171-L177), it looks like if you set a negative time `screenCachedExpired` will always be `true` (because `lastModifiedInterval` should always be non-negative), which is the same as having no cache.

So technically, unless I misunderstand how this works, the actual truth is that Senna will cache if the value is greater than `0`, not `-1`.

🤷

julien: Oh ... right.

julien: I'll fix that

matuzalemsteles:
I'm not sure if I managed to give a clear understanding of how Senna.js works, but let me know if we can add more things, the idea is to pass on an understanding and not go into the methods so much. Does that make sense?

wincent:
> I'm not sure if I managed to give a clear understanding of how Senna.js works, but let me know if we can add more things, the idea is to pass on an understanding and not go into the methods so much. Does that make sense?

I think it makes sense. The idea is not to explain every detail of the source code, but to give a high-level idea so that when somebody comes to look at the source, they have a "conceptual map" to find their way around. Without the "map", you have to reverse-engineer the thing in order to begin understanding it, and that's hard.

matuzalemsteles:
> I think it makes sense. The idea is not to explain every detail of the source code, but to give a high-level idea so that when somebody comes to look at the source, they have a "conceptual map" to find their way around. Without the "map", you have to reverse-engineer the thing in order to begin understanding it, and that's hard.

I completely agree 🙂.

dxp/sennajs_integration.md
```diff
@@ -0,0 +1,117 @@
...
+![](../images/sennajs-flip.png)
+
+Resource tracking happens when navigation is being made, this includes the `style` and `script` tags, Screen looks for those elements that are marked with the `data-senna-track` attribute within the content that is in a virtual document or in the DOM.
+
+There are two types of tracking for these attributes:
+
+-   **Permanent** These are resources that persist during navigation regardless of the route
+-   **Temporary** These are resources that are removed in the next route navigation
+
+> Any resource that has this attribute declared and is within the document's `body` will be moved to the document's `head`.
```
jbalsas, line 115: How are elements without the `data-senna-track` attribute treated by default? ;)

dxp/sennajs_integration.md
```diff
@@ -0,0 +1,117 @@
...
+
+![Simplified overview of Senna.js](../images/sennajs-overview.png)
+
+Senna.js, introduces the following concepts:
+
+-   **Routes** Determine which path will be controlled by Senna.js and which Screen will handle the content.
+-   **Surface** A DOM element with an `id` attribute, which is where the content will be rendered.
+-   **Screen** A special type of route handler to handle the navigation request.
+
+To simulate a single page application, Senna.js [tracks all links](https://github.com/liferay/senna.js/blob/059022b737de0f0108b7ef3f06f14d7ade3baa94/src/app/App.js#L104) within the Surface. When a link is triggered, Senna.js controls the event by preventing the default behavior to prevent navigation from being made by the browser and then triggers the navigation process.
```
jbalsas, line 72: I don't see this explained anywhere... how many surfaces do we have in DXP? which one(s)? This is interesting because people might thing this is working in a way when it's working in a different way.

jbalsas: Once you get that, I think it's also interesting to add some diagram or explanation of what type of content are we getting back from DXP backend and what do we do with it.

Spoiler alert, we get full documents as if it were full refreshes (not partials), and then there's some things that get filtered or ignored... I think this would be interesting to write down.

dxp/sennajs_integration.md
julien: I think it's "permanent" but I'll look into the code and document it

matuzalemsteles: Yeah, if the element is in the `head` without the attribute it will be permanent Senna.js will not touch this element. But if the element is in `body` or specifically within the content of the page without the attribute, the behavior will be like a temporary.

dxp/sennajs_integration.md
matuzalemsteles: Yeah, I think that in the concept of Senna.js there was to be more than one Surface but for DXP only one is used during startup, I don't know exactly why. But yes I can document these details more on the DXP side, in fact, it will happen independently of DXP as well.

jbalsas: > Yeah, I think that in the concept of Senna.js there was to be more than one Surface but for DXP only one is used during startup, I don't know exactly why.

I only vaguely remember why now... maybe it can be found in DXP's git history, I'll try to see if I can find the specific commits tomorrow and if not, I'll simply share what I remember.

matuzalemsteles: > I only vaguely remember why now... maybe it can be found in DXP's git history, I'll try to see if I can find the specific commits tomorrow and if not, I'll simply share what I remember.

Thanks, I'll try to look for something in the commits too.

> Spoiler alert, we get full documents as if it were full refreshes (not partials), and then there's some things that get filtered or ignored... I think this would be interesting to write down.

Yeah, this is an important thing, I'll make that clearer.

dxp/sennajs_integration.md
julien: Thanks @matuzalemsteles I'll add that right now

dxp/sennajs_integration.md
julien: The only surface I found was `LiferaySurface` but I'll also search in the git history

jbalsas: `LiferaySurface` is a class. When the code run, we instantiate and register a number of surfaces which is what's in the diagram above. Don't go looking in the history for code, figure out which surfaces we create and manage in runtime and then check if we did anything different at any other point in time.

julien: I probably should have said that `LiferaySurface` is a class, but the only instance of that class being created is [here](https://github.com/liferay/liferay-portal/blob/3ec7c6e5d08fa594303f140c55407f7158ecccb3/modules/apps/frontend-js/frontend-js-spa-web/src/main/resources/META-INF/resources/liferay/app/App.es.js#L79). 
At one point we changed [this](https://github.com/liferay/liferay-portal/commit/70b1010a4a521b5a4f955ae281c9e5019faa643a) to prevent overwritting the `document.body.id` if it has one.

matuzalemsteles:
I added some more information, let me know if I forgot any details.

dxp/sennajs_integration.md
```diff
@@ -59,6 +59,22 @@ These paths, are visible in the global `Liferay.SPA.excludedPaths` variable.
...
+
+-   `beforeNavigate`
+-   `beforeScreenFlip`
+-   `endNavigate`
+-   `screenActivate`
+-   `screenDeactivate`
+-   `screenDispose`
+-   `screenFlip`
+-   `screenLoad`
+-   `startNavigate`
```
wincent, line 74: I'd be interested in knowing some more about what each of these does.

For example, without looking at the code, I don't know the difference between `beforeNavigate`, `startNavigate` and `beforeScreenFlip` (although I can make a guess, and make assumptions about the order they fire in).

Likewise, I don't know the order of or difference between `screenDeactivate` and `screenDispose` — again, I can take a guess, or look at the code, but it would be cool if it were spelled out for me.

Or another way to think of it, we might want to define what the basic terms mean:

- What "flipping" a screen means.
- What "activating" and "deactivating" (and "disposing") a screen means.

Of the terms, the only ones that have relatively clear meanings are "navigate" and "load", but even then we need to be clear about what they mean in the context of Senna.

wincent (commented):
I don't want to let "perfect be the enemy of the good" here, so I think we should ship this soon. There is just one area I've commented on where I think there is a bit of a gap; I leave it up to you @julien and @matuzalemsteles to decide whether it is worth filling that gap.

dxp/sennajs_integration.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Form submission
+
+The form submission will also be handled by the Senna.js `App` instance, unless the form's `action` attribute is set to `get`
+
+## Block list
+
+It's also possible to tell the Senna.js `App` instance which portlets it needs to ignore, in which case it will not be used for navigation or form submission.
+
+Any portlet that is set to not use "Single Application Mode", or any portlet that is not deployed (and not ready) is automatically added to this list.
```
jbalsas, line 40: Might be good to explain how this is done technically... via the OSGi header or by other means.

dxp/sennajs_integration.md
```diff
@@ -0,0 +1,145 @@
...
+By default DXP sets this value to `-1` which means that the cache is disabled, if you want Senna.js cache to never be cleared you can set this value to `0`
+or specify the value (in minutes) after when Senna.js cache should be cleared.
+
+The cache expiration time can also be configured by navigating to **Control Panel** > **System Settings** > **Infrastructure** > **Frontend SPA Infrastructure** and changing the value specified in the **Cache Expiration Time** field.
+
+With caching enabled, once a resource is loaded, it's content is cached in memory and is retrieved later on without any additional request.
+
+## Excluded paths
+
+The `frontend-js-spa-web` module also introduces a property called [`spa.exluced.paths`](https://github.com/liferay/liferay-portal/blob/454d7f6a439dc668a0ce0529cff2cae87ee9a0d5/modules/apps/frontend-js/frontend-js-spa-web/src/main/resources/spa.properties) which, as its name indicates, is a list of paths that should be ignored by the `Senna.js` App instance.
```
jbalsas, line 54: I think there's a typo in `spa.exluced.paths`. Hope it's `excluded` and we didn't miss it the whole time :D

dxp/sennajs_integration.md
jbalsas: Agreed... a high level description of the lifecycle events would be helpful. We have some information in the source code itself or in sennajs.com, so just putting together some of it here should be enough, I think.

jbalsas (approved):
Other than the request to expand a little bit on the meaning of the lifecycle methods, this looks good to me! 😍

dxp/sennajs_integration.md
julien: :+1: good catch

dxp/sennajs_integration.md
```diff
@@ -84,30 +84,30 @@ By default, Senna.js dispatches some events during the navigation process that c
...
 
+-   `startNavigate`
+    Fired when navigation begins
+
 -   `beforeScreenFlip`
     Dispatched before a screen is visible
 
--   `endNavigate`
-    Dispatched after the content has been retrieved and inserted onto the page
+-   `screenLoad`
```
wincent, line 93: Is it true that `beforeScreenFlip` comes before `screenLoad`?

dxp/sennajs_integration.md
```diff
@@ -45,17 +45,17 @@ To disable single application mode in your portlet, you can follow these steps:
...
 
-    ```xml
-    <single-page-application>false</single-page-application>
-    ```
+        ```xml
+        <single-page-application>false</single-page-application>
+        ```
 
-    Alternatively, you can override the `isSinglePageApplication` method of the portlet to return false.
+        Alternatively, you can override the `isSinglePageApplication` method of the portlet to return false.
```
wincent: I think it was right before:

Screenshot 2020-10-05 -170739-PAc7RTj5@2x

This commit makes it look like this:

Screenshot 2020-10-05 -170704-7sFIeWNz@2x

(14 commits and counting; we had better squash all this before merging.)

julien: I'm going to fix this and squash, I don't know if it's me but each time I try to view the change on GitHub it's messed up

dxp/sennajs_integration.md
julien: no, it's not. I fixed that in https://github.com/liferay/liferay-frontend-guidelines/pull/223/commits/2e2b34b5b3726a1182a4a1b16801bf2977b29771

wincent (approved):
Seems everybody is happy with this, and the rendered output looks good, so let's get this in.

PR #226, opened 2020-10-02
docs: Adds Demystifying `aui:script` and `liferay-frontend:component` tags docs

### LINKED ISSUE #225: Document aui:script and liferay-frontend:component tag
`aui:script` tag is widely used on DXP

It can be cool document what this tag does and usage.

Some cool things that can be explained: 

* This taglib Internals, what it does
* `require` property
* `sandbox`property

diegonvs:
Closes #225

### DIFF
--- a/dxp/aui_script.md
+++ b/dxp/aui_script.md
@@ -12,7 +12,7 @@ We have three possible use cases when using this tag:
 
 1. Inline JS on the JSP
 2. Import an AUI/YUI module
-3. Import a ES module
+3. Import an ES module
 
 #### Inline JS on the JSP
 
@@ -74,30 +74,32 @@ AUI().use('liferay-alert', function (A) {
 }
 ```
 
-Note in the AlloyUI components, coming from the alloy-ui repository, classes are made available from the global variable `A` on the scope of the `aui:script`. Here as in this [code snippet the Menu component is defined](https://github.com/liferay/alloy-ui/blob/master/src/aui-menu/js/aui-menu.js#L39]).
+Note in the AlloyUI components, coming from the alloy-ui repository, classes are made available from the global variable `A` on the scope of the `aui:script`(as in this [code snippet the Menu component is defined](https://github.com/liferay/alloy-ui/blob/master/src/aui-menu/js/aui-menu.js#L39)).
 
-However, components made available from the `frontend-js-aui-web` module will be under the global Liferay variable. Like this [example here, `Liferay.Upload`](https://github.com/liferay/liferay-portal/blob/815f48f484351e18b61e4b9c9fbf40f0609bdc56/modules/apps/frontend-js/frontend-js-aui-web/src/main/resources/META-INF/resources/liferay/upload.js#L1541), used in the code example above.
+However, components made available from the `frontend-js-aui-web` module will be under the global Liferay variable. Like this [example here, where `Liferay.Upload` is implemented](https://github.com/liferay/liferay-portal/blob/815f48f484351e18b61e4b9c9fbf40f0609bdc56/modules/apps/frontend-js/frontend-js-aui-web/src/main/resources/META-INF/resources/liferay/upload.js#L1541), used in the code example above.
 
-#### Import a ES module using `require` property
+#### Import an ES module using `require` property
 
 Allows users to require a modern ES module as in the following example:
 
-Considering We already have `metal-clipboard` and `metal-dom` dependencies listed [on our module package.json](https://help.liferay.com/hc/en-us/articles/360018159771-The-Structure-of-OSGi-Bundles-Containing-npm-Packages-)
+Considering we already have `metal-clipboard` and `metal-dom` dependencies listed [in our module package.json](https://help.liferay.com/hc/en-us/articles/360018159771-The-Structure-of-OSGi-Bundles-Containing-npm-Packages-)
 
-By default the name of the loaded resource will be the transformed path to camel-case default exported module/file:
+By default, the name of the variable referring to the required resource will be composed by transforming the resource's path to camel case. For example:
 
 ```jsp
 <aui:script require="metal-clipboard/src/Clipboard">
-    // Note We have the provided path to require metal-clipboard/src/Clipboard normalized and camel-case as a variable name
+    // Note how the required metal-clipboard/src/Clipboard is made available in the metalClipboardSrcClipboard variable
 	new metalClipboardSrcClipboard.default();
 </aui:script>
 ```
 
-Also, We can set an alias with the `as` word after the provided path. It will be look like:
+> You can have a look at the details of the algorithm for composing variable names in the [VariableUtil.generateVariable()](https://github.com/liferay/liferay-portal/blob/b69a6cc79c9cd2d3e13ef2879c6f158be0b1c094/portal-kernel/src/com/liferay/portal/kernel/servlet/taglib/aui/VariableUtil.java#L33) method.
+
+Also, We can set an alias with the `as` word after the provided path. It would look like:
 
 ```jsp
 <aui:script require="my-custom-lib/src/path/main as myLib">
-	// ... code that use some function from myLib like `myLib.someFunction();`
+	// ... code that uses some function from myLib like `myLib.someFunction();`
 </aui:script>
 ```
 
@@ -109,7 +111,7 @@ Liferay.Loader.require('my-custom-lib/src/path/main', function(myCustomLibSrcPat
 		(function() {
 			var myLib = myCustomLibSrcPathMain;
 			var $ = AUI.$;var _ = AUI._;
-			// ... code that use some function from myLib like `myLib.someFunction();`
+			// ... code that uses some function from myLib like `myLib.someFunction();`
 		})();
 	} catch (err) {
 		console.error(err);
@@ -121,9 +123,9 @@ You can find more information on [this link](https://help.liferay.com/hc/en-us/a
 
 #### Using `position` to control script placement
 
-This property will change the way in which / how the script will be injected into the page. By default, those scripts that are wrapped by `aui:script` will be added to the minified files as described [here](#inline-js-on-the-jsp).
+This property will change the way in which the script will be injected into the page. By default, those scripts that are wrapped by `aui:script` will be added to the minified files as described in ["Inline JS on the JSP"](#inline-js-on-the-jsp).
 
-When placing the option of `inline` for this property, the scripts will be placed in a script tag obeying the order of the DOM where the JSP is being rendered. See [here](https://github.com/liferay/liferay-portal/blob/815f48f484351e18b61e4b9c9fbf40f0609bdc56/util-taglib/src/com/liferay/taglib/aui/ScriptTag.java#L143..L165).
+When placing the option of `inline` for this property, the scripts will be placed in a script tag obeying the order of the DOM where the JSP is being rendered. See the [aui:script tag implementation here.](https://github.com/liferay/liferay-portal/blob/815f48f484351e18b61e4b9c9fbf40f0609bdc56/util-taglib/src/com/liferay/taglib/aui/ScriptTag.java#L143..L165).
 
 #### `sandbox`
 
@@ -143,6 +145,6 @@ As a short brief, it will limit the scope of the script tag. See the following e
 
 #### Caveats
 
-You can't use both require and use in the same tag. use is for declaring dependencies on AUI modules, while require gives access to ES modules.
+You can't use both `require` and `use` in the same tag. use is for declaring dependencies on AUI modules, while require gives access to ES modules.
 
-You can see more information about `aui:script` [here](https://help.liferay.com/hc/en-us/articles/360017882752-Loading-Modules-with-AUI-Script-in-Liferay-DXP).
+You can see more information about `aui:script` in ["Loading Modules with AUI Script in Liferay DXP"](https://help.liferay.com/hc/en-us/articles/360017882752-Loading-Modules-with-AUI-Script-in-Liferay-DXP).
--- a/dxp/aui_script.md
+++ b/dxp/aui_script.md
@@ -1,12 +1,8 @@
-# Demystifying `aui:script` and `liferay-frontend:component` tags
-
-In JSP files you may notice a huge number of `aui:script` and `liferay-frontend:component` tags.
+# Demystifying `aui:script` tag
 
 ## IMPORTANT
 
 Infrastructure is sunsetting AlloyUI since when ["The status and direction of the frontend infrastructure in Liferay 7 & DXP"](https://liferay.dev/blogs/-/blogs/the-status-and-direction-of-the-frontend-infrastructure-in-liferay-7-dxp) blog post was released. Also, We are working hard in removing those components on [Remove AUI / YUI from Liferay Portal](https://issues.liferay.com/browse/LPS-98564) Epic. Please, avoid using AlloyUI JS components and tags, except for `aui:script` tag.
-
-## `aui:script` tag
 
 ### What it does
 
@@ -150,32 +146,3 @@ As a short brief, it will limit the scope of the script tag. See the following e
 You can't use both require and use in the same tag. use is for declaring dependencies on AUI modules, while require gives access to ES modules.
 
 You can see more information about `aui:script` [here](https://help.liferay.com/hc/en-us/articles/360017882752-Loading-Modules-with-AUI-Script-in-Liferay-DXP).
-
-## `liferay-frontend:component` tag
-
-We can use this tag for removing JavaScript code from JSPs. This tag accepts a `module` property through which We can pass a relative path for the script to be loaded. Also, this tag registers automatically the component to [Liferay’s component global registry](https://github.com/liferay/liferay-frontend-guidelines/blob/bc6dae8514af04a6384a7bea9d4ddf266087ffc5/dxp/liferay_component.md#register), depending on the `componentId` property. When using this tag, `namespace`(or `portletNamespace`) and `spritemap` props will become available.
-
-For example, in the JSP page:
-
-```jsp
-...
-<liferay-frontend:component
-	componentId="myComponent"
-	module="path/to/my/js/myComponent.js"
-/>
-...
-```
-
-And when exporting the `myComponent.js` file We can define a default function there like:
-
-```js
-export default function myComponent({namespace, spritemap, ...props}) {
-	// ... code that uses namespace, spritemap etc
-}
-```
-
-This tag can be very handy if you want to use modern JS features, since We don’t have support for transpiling modern JS from taglibs.
-
-Also, this tag adds some treatment for handling JavaScript in our JSPs like [adding special treatment for our built-in SPA framework](https://github.com/liferay/liferay-portal/blob/815f48f484351e18b61e4b9c9fbf40f0609bdc56/modules/apps/frontend-taglib/frontend-taglib/src/main/java/com/liferay/frontend/taglib/servlet/taglib/ComponentTag.java#L225).
-
-All the content will be injected on the page using `AUI_SCRIPT_DATA` already explained [here](https://github.com/liferay/liferay-frontend-guidelines/blob/bc6dae8514af04a6384a7bea9d4ddf266087ffc5/dxp/resource_injection.md#scriptdata).
--- a/dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
+++ b/dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
@@ -1,0 +1,114 @@
+# Demystifying `aui:script` and `liferay-frontend:component` tags
+
+On JSP context you may notice a huge number of `aui:script` and `liferay-frontend:component` tags.
+
+## `aui:script` tag
+
+### What it does:
+
+This taglib allow users to use and require another JS module/resources inside a JSP.
+
+This taglib have three possible use cases:
+
+1. Pure JS on the JSP
+2. Import a AUI/YUI module
+3. Import a ES6 module
+
+#### Pure JS on the JSP
+
+By just placing JavaScript code inside an `aui:script` tag, you can take advantage of optimizing scripts and place them to be made available in the final compressed JavaScript bundle, see [combo for more details](https: / /github.com/liferay/liferay-frontend-guidelines/blob/45bad0334cb467e45e68aee322a2a4bc0b6979d3/dxp/javaScript_minification.md#comboservlet). It is worth mentioning that the same rule applies for the following use cases.
+
+#### Import an AUI/YUI module using `use` property
+
+When using the `use` property, the AUI / YUI module listed in the` use` list will be imported to the script data. If you want to know a little more about ScriptData, see this [link on how resources are injected into DXP](https://github.com/liferay/liferay-frontend-guidelines/blob/ea185686db8d562ac11cc2b95d2e9e7d7a0c6547/dxp/resource_injection.md#scriptdata).
+
+You can import several YUI / AUI modules within the same `use` property, as in this example:
+
+```jsp
+<aui:script use="liferay-upload">
+...
+        var liferayUpload = new Liferay.Upload({
+...
+</aui:script>
+```
+
+Note in the AlloyUI components, coming from the alloy-ui repository, classes are made available from the global variable `A`, as in this [code snippet the Menu component is defined](https://github.com/liferay/alloy-ui/blob/master/src/aui-menu/js/aui-menu.js#L39]).
+
+However, components made available from the `frontend-js-aui-web` module will be under the global Liferay variable. Like this [example here, `Liferay.Upload`](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-js/frontend-js-aui-web/src/main/resources/META-INF/resources/liferay/upload.js#L1541), used in the code example above.
+
+#### Import an ES6 module using `require` property
+
+Allows users to require a ES6 module as the following example:
+
+Considering that we already have `metal-clipboard` and `metal-dom` dependencies listed [on our module package.json](https://help.liferay.com/hc/en-us/articles/360018159771-The-Structure-of-OSGi-Bundles-Containing-npm-Packages-), We have:
+
+```jsp
+<aui:script require="metal-clipboard/src/Clipboard">
+    new metalClipboardSrcClipboard.default();
+</aui:script>
+```
+
+Also, we can set an alias with the `as` word after the provided path. It will be looking like:
+
+```jsp
+<aui:script require="metal-dom/all/dom.js as dom">
+
+	function myFunction(element) {
+		dom.addClasses(element, 'hide');
+	}
+
+	Liferay.once('something', myFunction);
+
+</aui:script>
+```
+
+#### Settling the way which the script will be placed on the page using `position`:
+
+This property will change the way in which / how the script will be injected into the page. By default, those scripts that are wrapped by `aui:script` will be added to the minified files as described here (Pure JS on the JSP [PUT THE LINK HERE OH]).
+
+When placing the option of `inline` for this property, the scripts will be placed in a script tag obeying the order of the DOM where the JSP is being rendered. See [here](https://github.com/liferay/liferay-portal/blob/master/util-taglib/src/com/liferay/taglib/aui/ScriptTag.java#L143..L165).
+
+#### `sandbox`:
+
+The purpose for the sandbox attribute is to wrap the contents of the script tag in a self-invoking function, and to enable short-handing the `$` and `_` variables rather than calling `AUI.$` and `AUI._`.
+
+#### Caveats:
+
+We couldn’t use both `require` and `use` in the same file. Because `use` property calls the modularization for alloy-ui modules and `require` follows ES6 modularized files.
+
+You can see more information about `aui:script` [here](https://help.liferay.com/hc/en-us/articles/360017882752-Loading-Modules-with-AUI-Script-in-Liferay-DXP).
+
+## `liferay-frontend:component` tag
+
+We can use this tag for removing JavaScript code from JSPs. This tag accepts a `module` property wether we can pass a relative path for the script to be loaded. Also, this tag registers automatically the component to [Liferay’s component global registry](https://github.com/liferay/liferay-frontend-guidelines/blob/master/dxp/liferay_component.md#register), depending on the `componentId` property. When using this tag, `namespace`(or `portletNamespace`) and `spritemap` props will turn available.
+
+Here is an example of usage of this tag:
+
+On the JSP context:
+
+```jsp
+
+...
+<liferay-frontend:component
+	componentId="myComponent"
+	module="path/to/my/js/myComponent.js"
+/>
+...
+```
+
+And when exporting the `myComponent.js` file we can define a default function there like:
+
+```js
+export default function myComponent(props) {
+	console.log(props.namespace);
+	// it will return the portlet namespace where the liferay-frontend:component tag were called
+	console.log(props.spritemap);
+	// it will return the path to the page's spritemap
+}
+```
+
+This taglib can be very handy if you want to use modern ES6 features, since We don’t have support for transpiling ES6 from taglibs.
+
+Also, this tag adds some treatments for handling JavaScript on our JSPs like [adding special treatments for our built-in SPA framework](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-taglib/frontend-taglib/src/main/java/com/liferay/frontend/taglib/servlet/taglib/ComponentTag.java#L225).
+
+All the content will be injected on the page using `AUI_SCRIPT_DATA` already explained [here](https://github.com/liferay/liferay-frontend-guidelines/blob/ea185686db8d562ac11cc2b95d2e9e7d7a0c6547/dxp/resource_injection.md#scriptdata).
--- a/dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
+++ b/dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
@@ -1,93 +1,163 @@
 # Demystifying `aui:script` and `liferay-frontend:component` tags
 
-On JSP context you may notice a huge number of `aui:script` and `liferay-frontend:component` tags.
+In JSP files you may notice a huge number of `aui:script` and `liferay-frontend:component` tags.
+
+## IMPORTANT
+
+Infrastructure is sunsetting AlloyUI since when ["The status and direction of the frontend infrastructure in Liferay 7 & DXP"](https://liferay.dev/blogs/-/blogs/the-status-and-direction-of-the-frontend-infrastructure-in-liferay-7-dxp) blog post was released. Also, We are working hard in removing those components on [Remove AUI / YUI from Liferay Portal](https://issues.liferay.com/browse/LPS-98564) Epic. Please, avoid using AlloyUI JS components and tags, except for `aui:script` tag.
 
 ## `aui:script` tag
 
-### What it does:
+### What it does
 
-This taglib allow users to use and require another JS module/resources inside a JSP.
+This tag allow users to use and require ES module/resources inside a JSP.
 
-This taglib have three possible use cases:
+We have three possible use cases when using this tag:
 
-1. Pure JS on the JSP
-2. Import a AUI/YUI module
-3. Import a ES6 module
+1. Inline JS on the JSP
+2. Import an AUI/YUI module
+3. Import a ES module
 
-#### Pure JS on the JSP
+#### Inline JS on the JSP
 
-By just placing JavaScript code inside an `aui:script` tag, you can take advantage of optimizing scripts and place them to be made available in the final compressed JavaScript bundle, see [combo for more details](https: / /github.com/liferay/liferay-frontend-guidelines/blob/45bad0334cb467e45e68aee322a2a4bc0b6979d3/dxp/javaScript_minification.md#comboservlet). It is worth mentioning that the same rule applies for the following use cases.
+By just placing JavaScript code inside an `aui:script` tag, you can take advantage of optimizing scripts and place them to be made available in the final compressed JavaScript bundle. It can be achieved when appending the script to the [`AUI_SCRIPT_DATA` WebKey](https://github.com/liferay/liferay-portal/blob/67b569099146a4b999e2fad7d7d1a9794a337f0b/portal-kernel/src/com/liferay/portal/kernel/util/WebKeys.java#L55) and all these scripts will be applied at the [bottom of the page](https://github.com/liferay/liferay-portal/blob/e3ffac158e0ec5acc5c67069fbd7ba688d3c78d4/portal-web/docroot/html/common/themes/bottom.jsp#L44). Finally the [Aggregate Filter will take care](https://github.com/liferay/liferay-frontend-guidelines/blob/bc6dae8514af04a6384a7bea9d4ddf266087ffc5/dxp/resource_serving.md#aggregate-filter), of creating the final bundle with all the JavaScript on the page. It is worth mentioning that the same rule applies to all the following use cases.
 
-#### Import an AUI/YUI module using `use` property
+Like this example:
 
-When using the `use` property, the AUI / YUI module listed in the` use` list will be imported to the script data. If you want to know a little more about ScriptData, see this [link on how resources are injected into DXP](https://github.com/liferay/liferay-frontend-guidelines/blob/ea185686db8d562ac11cc2b95d2e9e7d7a0c6547/dxp/resource_injection.md#scriptdata).
+```jsp
+<aui:script>
+	Liferay.fire('myGlobalEvent', {
+		payload: 'myPayload',
+	});
+</aui:script>
+```
+
+it will be turned on:
+
+```js
+Liferay.fire('myGlobalEvent', {
+	payload: 'myPayload'
+});
+```
+
+#### Import an AUI/YUI module using the `use` property
+
+> ⚠️ This is a deprecated use case, though it still works if someone configures it.⚠️
+
+Each AUI / YUI module listed as a value for the `use` property will be [loaded asyncronously by the `AUI().use` function](https://github.com/yui/yui3/blob/master/src/yui/js/yui.js#L1323..L1331).
 
 You can import several YUI / AUI modules within the same `use` property, as in this example:
 
+Here is an example:
+
 ```jsp
-<aui:script use="liferay-upload">
-...
-        var liferayUpload = new Liferay.Upload({
-...
+<aui:script use="liferay-alert">
+	new Liferay.Alert(
+		{
+			closeable: true,
+			message: 'I like drink tea while eating french potatoes',
+			type: 'success'
+		}
+	).render('#myId');
 </aui:script>
 ```
 
-Note in the AlloyUI components, coming from the alloy-ui repository, classes are made available from the global variable `A`, as in this [code snippet the Menu component is defined](https://github.com/liferay/alloy-ui/blob/master/src/aui-menu/js/aui-menu.js#L39]).
+The output will be a code like this:
 
-However, components made available from the `frontend-js-aui-web` module will be under the global Liferay variable. Like this [example here, `Liferay.Upload`](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-js/frontend-js-aui-web/src/main/resources/META-INF/resources/liferay/upload.js#L1541), used in the code example above.
+```js
+AUI().use('liferay-alert', function (A) {
+	(function() {var $ = AUI.$;var _ = AUI._;
+		new Liferay.Alert(
+			{
+				closeable: true,
+				message: 'I like drink tea while eating french potatoes',
+				type: 'success'
+			}
+		).render('#myId');
+	})()
+}
+```
 
-#### Import an ES6 module using `require` property
+Note in the AlloyUI components, coming from the alloy-ui repository, classes are made available from the global variable `A` on the scope of the `aui:script`. Here as in this [code snippet the Menu component is defined](https://github.com/liferay/alloy-ui/blob/master/src/aui-menu/js/aui-menu.js#L39]).
 
-Allows users to require a ES6 module as the following example:
+However, components made available from the `frontend-js-aui-web` module will be under the global Liferay variable. Like this [example here, `Liferay.Upload`](https://github.com/liferay/liferay-portal/blob/815f48f484351e18b61e4b9c9fbf40f0609bdc56/modules/apps/frontend-js/frontend-js-aui-web/src/main/resources/META-INF/resources/liferay/upload.js#L1541), used in the code example above.
 
-Considering that we already have `metal-clipboard` and `metal-dom` dependencies listed [on our module package.json](https://help.liferay.com/hc/en-us/articles/360018159771-The-Structure-of-OSGi-Bundles-Containing-npm-Packages-), We have:
+#### Import a ES module using `require` property
+
+Allows users to require a modern ES module as in the following example:
+
+Considering We already have `metal-clipboard` and `metal-dom` dependencies listed [on our module package.json](https://help.liferay.com/hc/en-us/articles/360018159771-The-Structure-of-OSGi-Bundles-Containing-npm-Packages-)
+
+By default the name of the loaded resource will be the transformed path to camel-case default exported module/file:
 
 ```jsp
 <aui:script require="metal-clipboard/src/Clipboard">
-    new metalClipboardSrcClipboard.default();
+    // Note We have the provided path to require metal-clipboard/src/Clipboard normalized and camel-case as a variable name
+	new metalClipboardSrcClipboard.default();
 </aui:script>
 ```
 
-Also, we can set an alias with the `as` word after the provided path. It will be looking like:
+Also, We can set an alias with the `as` word after the provided path. It will be look like:
 
 ```jsp
-<aui:script require="metal-dom/all/dom.js as dom">
-
-	function myFunction(element) {
-		dom.addClasses(element, 'hide');
-	}
-
-	Liferay.once('something', myFunction);
-
+<aui:script require="my-custom-lib/src/path/main as myLib">
+	// ... code that use some function from myLib like `myLib.someFunction();`
 </aui:script>
 ```
 
-#### Settling the way which the script will be placed on the page using `position`:
+The output will be:
 
-This property will change the way in which / how the script will be injected into the page. By default, those scripts that are wrapped by `aui:script` will be added to the minified files as described here (Pure JS on the JSP [PUT THE LINK HERE OH]).
+```js
+Liferay.Loader.require('my-custom-lib/src/path/main', function(myCustomLibSrcPathMain) {
+	try {
+		(function() {
+			var myLib = myCustomLibSrcPathMain;
+			var $ = AUI.$;var _ = AUI._;
+			// ... code that use some function from myLib like `myLib.someFunction();`
+		})();
+	} catch (err) {
+		console.error(err);
+	}
+}
+```
 
-When placing the option of `inline` for this property, the scripts will be placed in a script tag obeying the order of the DOM where the JSP is being rendered. See [here](https://github.com/liferay/liferay-portal/blob/master/util-taglib/src/com/liferay/taglib/aui/ScriptTag.java#L143..L165).
+You can find more information on [this link](https://help.liferay.com/hc/en-us/articles/360017882752-Loading-Modules-with-AUI-Script-in-Liferay-DXP#loading-es2015-and-metaljs-modules)
 
-#### `sandbox`:
+#### Using `position` to control script placement
 
-The purpose for the sandbox attribute is to wrap the contents of the script tag in a self-invoking function, and to enable short-handing the `$` and `_` variables rather than calling `AUI.$` and `AUI._`.
+This property will change the way in which / how the script will be injected into the page. By default, those scripts that are wrapped by `aui:script` will be added to the minified files as described [here](#inline-js-on-the-jsp).
 
-#### Caveats:
+When placing the option of `inline` for this property, the scripts will be placed in a script tag obeying the order of the DOM where the JSP is being rendered. See [here](https://github.com/liferay/liferay-portal/blob/815f48f484351e18b61e4b9c9fbf40f0609bdc56/util-taglib/src/com/liferay/taglib/aui/ScriptTag.java#L143..L165).
 
-We couldn’t use both `require` and `use` in the same file. Because `use` property calls the modularization for alloy-ui modules and `require` follows ES6 modularized files.
+#### `sandbox`
+
+The purpose for the sandbox attribute is to wrap the contents of the script tag in an [Immediately Invoked Function Expression/IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE).
+
+As a short brief, it will limit the scope of the script tag. See the following example:
+
+```jsp
+<aui:script>
+   var myVar = 'foo'; // This variable will be available globally
+</aui:script>
+
+<aui:script sandbox="<%= true %>>
+   var myVar = 'foo'; // This variable will NOT be available globally
+</aui:script>
+```
+
+#### Caveats
+
+You can't use both require and use in the same tag. use is for declaring dependencies on AUI modules, while require gives access to ES modules.
 
 You can see more information about `aui:script` [here](https://help.liferay.com/hc/en-us/articles/360017882752-Loading-Modules-with-AUI-Script-in-Liferay-DXP).
 
 ## `liferay-frontend:component` tag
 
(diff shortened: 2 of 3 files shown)


dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+#### Settling the way which the script will be placed on the page using `position`:
+
+This property will change the way in which / how the script will be injected into the page. By default, those scripts that are wrapped by `aui:script` will be added to the minified files as described here (Pure JS on the JSP [PUT THE LINK HERE OH]).
+
+When placing the option of `inline` for this property, the scripts will be placed in a script tag obeying the order of the DOM where the JSP is being rendered. See [here](https://github.com/liferay/liferay-portal/blob/master/util-taglib/src/com/liferay/taglib/aui/ScriptTag.java#L143..L165).
+
+#### `sandbox`:
+
+The purpose for the sandbox attribute is to wrap the contents of the script tag in a self-invoking function, and to enable short-handing the `$` and `_` variables rather than calling `AUI.$` and `AUI._`.
```
diegonvs, line 73: Two questions here:

1. Taking a look on the implementation, We aren't checking if JQuery is available on the page or not. Should I open a ticket for it?

2. What are the benefits in wrapping a script on an anonymous function?

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
+# Demystifying `aui:script` and `liferay-frontend:component` tags
+
+On JSP context you may notice a huge number of `aui:script` and `liferay-frontend:component` tags.
+
+## `aui:script` tag
```
diegonvs, line 5: Should I mention We are moving from using using JS code within our JSPs?

https://github.com/liferay-frontend/liferay-portal/pull/407

diegonvs:
@izaera, please correct me if I'm wrong somewhere on this file about modularization :p

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+### What it does:
+
+This taglib allow users to use and require another JS module/resources inside a JSP.
```
carloslancha, line 9: It's a tag, not a taglib (group of tags) ;)

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+This taglib have three possible use cases:
```
carloslancha, line 11: taglib -> tag

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+1. Pure JS on the JSP
+2. Import a AUI/YUI module
+3. Import a ES6 module
+
+#### Pure JS on the JSP
```
carloslancha, line 17: I'd add an example to this use case

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+However, components made available from the `frontend-js-aui-web` module will be under the global Liferay variable. Like this [example here, `Liferay.Upload`](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-js/frontend-js-aui-web/src/main/resources/META-INF/resources/liferay/upload.js#L1541), used in the code example above.
+
+#### Import an ES6 module using `require` property
+
+Allows users to require a ES6 module as the following example:
+
+Considering that we already have `metal-clipboard` and `metal-dom` dependencies listed [on our module package.json](https://help.liferay.com/hc/en-us/articles/360018159771-The-Structure-of-OSGi-Bundles-Containing-npm-Packages-), We have:
+
+```jsp
+<aui:script require="metal-clipboard/src/Clipboard">
```
carloslancha, line 46: Worth explaining that the name of the component variable is set converting the path to camelCase?

`metal-clipboard/src/Clipboard` -> `metalClipboardSrcClipboard`

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	Liferay.once('something', myFunction);
+
+</aui:script>
+```
+
+#### Settling the way which the script will be placed on the page using `position`:
+
+This property will change the way in which / how the script will be injected into the page. By default, those scripts that are wrapped by `aui:script` will be added to the minified files as described here (Pure JS on the JSP [PUT THE LINK HERE OH]).
```
carloslancha, line 67: `[PUT THE LINK HERE OH]` ;)

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
carloslancha: 2. Wrapping it you're sandboxing it, which means that you are limiting the scope. i.e:
```
<aui:script>
   var myVar = 'foo'; // This variable will be available globally
</aui:script>

<aui:script sandbox="<%= true %>>
   var myVar = 'foo'; // This variable will NOT be available globally
</aui:script>

File: dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md [hunk 8]
```diff
@@ -0,0 +1,114 @@
...
+```js
+export default function myComponent(props) {
+	console.log(props.namespace);
+	// it will return the portlet namespace where the liferay-frontend:component tag were called
+	console.log(props.spritemap);
+	// it will return the path to the page's spritemap
+}
+```
+
+This taglib can be very handy if you want to use modern ES6 features, since We don’t have support for transpiling ES6 from taglibs.
```
carloslancha, line 110: taglib -> tag

carloslancha (commented):
Adding info about the `defaultEventHandler` API, what it is and how to use it to `liferay-frontend:component` may be useful too.

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+This taglib have three possible use cases:
+
+1. Pure JS on the JSP
+2. Import a AUI/YUI module
+3. Import a ES6 module
+
+#### Pure JS on the JSP
+
+By just placing JavaScript code inside an `aui:script` tag, you can take advantage of optimizing scripts and place them to be made available in the final compressed JavaScript bundle, see [combo for more details](https: / /github.com/liferay/liferay-frontend-guidelines/blob/45bad0334cb467e45e68aee322a2a4bc0b6979d3/dxp/javaScript_minification.md#comboservlet). It is worth mentioning that the same rule applies for the following use cases.
```
wincent, line 19: Unless I am misunderstanding how this works, I don't think the combo servlet has much to do with this at all. `aui:script` tags don't have their JS extracted out in order to get served separately via the combo servlet; they get spat out into the HTML of the page markup itself.

Also, that link looks broken (spaces around the `/`, and the case of the `S` looks wrong).

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+#### Import an AUI/YUI module using `use` property
```
wincent, line 21: "using `use`" → "using the `use`"

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+When using the `use` property, the AUI / YUI module listed in the` use` list will be imported to the script data. If you want to know a little more about ScriptData, see this [link on how resources are injected into DXP](https://github.com/liferay/liferay-frontend-guidelines/blob/ea185686db8d562ac11cc2b95d2e9e7d7a0c6547/dxp/resource_injection.md#scriptdata).
```
wincent, line 23: I might be wrong about this, but it doesn't match my mental model of how things work. AUI modules are loaded asynchronously and not "imported to the script data" (that makes it sound like they are made available when the page loads).

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+```jsp
+<aui:script use="liferay-upload">
+...
+        var liferayUpload = new Liferay.Upload({
+...
+</aui:script>
+```
+
+Note in the AlloyUI components, coming from the alloy-ui repository, classes are made available from the global variable `A`, as in this [code snippet the Menu component is defined](https://github.com/liferay/alloy-ui/blob/master/src/aui-menu/js/aui-menu.js#L39]).
```
wincent, line 35: I'd say we can go pretty light on the AlloyUI documentation, and possibly even omit it. We don't want anybody writing new code that uses them (unlike `aui:script` itself, which we are happy to have people keep on using, for now).

In fact, I think this whole document needs a big warning at the top making it clear which parts of this API are non-deprecated and ok to use, and which parts are legacy and to be avoided.

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+However, components made available from the `frontend-js-aui-web` module will be under the global Liferay variable. Like this [example here, `Liferay.Upload`](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-js/frontend-js-aui-web/src/main/resources/META-INF/resources/liferay/upload.js#L1541), used in the code example above.
```
wincent, line 37: Is this true? I would have thought that at least some of them are not stuck onto the `Liferay` object, and are only loaded lazily on demand (via `use` etc) and never actually externalized on the global object.

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+#### Import an ES6 module using `require` property
```
wincent, line 39: As per above, say "ES module".

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+Allows users to require a ES6 module as the following example:
```
wincent, line 41: - "ES6 module" → "ES module"
- "as the following" → "as in the following"

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+Considering that we already have `metal-clipboard` and `metal-dom` dependencies listed [on our module package.json](https://help.liferay.com/hc/en-us/articles/360018159771-The-Structure-of-OSGi-Bundles-Containing-npm-Packages-), We have:
```
wincent, line 43: "We" → "we"

Also, this link seems totally random and unrelated.

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+```jsp
+<aui:script require="metal-clipboard/src/Clipboard">
+    new metalClipboardSrcClipboard.default();
+</aui:script>
+```
+
+Also, we can set an alias with the `as` word after the provided path. It will be looking like:
```
wincent, line 51: "It will be looking" → "It will look"

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+```jsp
+<aui:script require="metal-dom/all/dom.js as dom">
```
wincent, line 54: Missing from this is an explanation of _why_ you would want to use the `as` "operator", and guidance about what kind of alias to pick.

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+	function myFunction(element) {
+		dom.addClasses(element, 'hide');
+	}
+
+	Liferay.once('something', myFunction);
```
wincent, line 60: Delete all of this and replace it with:

```js
// ... code that uses the `dom` binding (eg. `dom.addClasses`) ...
```

The idea is to keep extraneous and distracting details out of the example (ie. what does this function have to do with it? What does `Liferay.once`?

I'm also a bit skeptical of the use of `metal-dom` for examples like this at all. I don't think we want to encourage the use of it when there are already built-in DOM APIs that do that kind of task (adding a `'hide'` class) without a dependency. And given that IE 11's days are numbered, there is even less reason to clutter our documentation with examples of legacy patterns that we don't want to encourage.

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+</aui:script>
+```
+
+#### Settling the way which the script will be placed on the page using `position`:
```
wincent, line 65: - "Settling" → "Setting"
- "the way which" → "the way in which"
- Drop the ":"

Or you might want to try an entirely different construction:

```
#### Using `position` to control script placement
```

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
wincent: "self-invoking function" is inaccurate. This would be a "self-invoking function" (ie. a recursive one):

```js
function fibonacci(n) {
    if (n <= 1) {
        return 1;
    }

    return fibonacci(n - 1) + fibonacci(n - 2);
}
```

The standard term that I think you're looking for is "IIFE". You don't even need to define it, just make it a link to the MDN [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) definition.

> Taking a look on the implementation, We aren't checking if JQuery is available on the page or not. Should I open a ticket for it?

I'd say absolutely not. This is a known issue, but we don't care: jQuery is off by default, and no code that we write from now on will ever rely on it.

So, I think you can drop the references to both `$` and `_`. We do not want to encourage or even admit that jQuery and Underscore exist.

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+This property will change the way in which / how the script will be injected into the page. By default, those scripts that are wrapped by `aui:script` will be added to the minified files as described here (Pure JS on the JSP [PUT THE LINK HERE OH]).
+
+When placing the option of `inline` for this property, the scripts will be placed in a script tag obeying the order of the DOM where the JSP is being rendered. See [here](https://github.com/liferay/liferay-portal/blob/master/util-taglib/src/com/liferay/taglib/aui/ScriptTag.java#L143..L165).
+
+#### `sandbox`:
```
wincent, line 71: Kill the ":".

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+The purpose for the sandbox attribute is to wrap the contents of the script tag in a self-invoking function, and to enable short-handing the `$` and `_` variables rather than calling `AUI.$` and `AUI._`.
+
+#### Caveats:
```
wincent, line 75: Delete ":"

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+We couldn’t use both `require` and `use` in the same file. Because `use` property calls the modularization for alloy-ui modules and `require` follows ES6 modularized files.
```
wincent, line 77: - "couldn't" → "can't"
- "ES6" → "ES"

Also, I think this is wrong, isn't it? You _can_ use them in the same _file_; you just can't use them in the same tag.

Maybe something like this:

> You can't use both `require` and `use` in the same tag. `use` is for declaring dependencies on AUI modules, while `require` gives access to ES modules.

Technically the thing you're requiring isn't actually an ES module at all. It may have started off as one in our codebase, but by the time it goes through the build process it's actually an AMD module, wrapped by the bundler. We might want to make that explicit somewhere up above, in one of the places where we first talk about `require` or about ES modules.

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+You can see more information about `aui:script` [here](https://help.liferay.com/hc/en-us/articles/360017882752-Loading-Modules-with-AUI-Script-in-Liferay-DXP).
+
+## `liferay-frontend:component` tag
+
+We can use this tag for removing JavaScript code from JSPs. This tag accepts a `module` property wether we can pass a relative path for the script to be loaded. Also, this tag registers automatically the component to [Liferay’s component global registry](https://github.com/liferay/liferay-frontend-guidelines/blob/master/dxp/liferay_component.md#register), depending on the `componentId` property. When using this tag, `namespace`(or `portletNamespace`) and `spritemap` props will turn available.
```
wincent, line 83: - "wether" → "whether", but I think you actually means something else? Like "in which" or "by which" or "through which" or "via which" (all of those work).
- "turn available" → "become available".

The fact that you are linking to the documentation about [`Liferay.Component`](https://github.com/liferay/liferay-frontend-guidelines/blob/master/dxp/liferay_component.md) makes me wonder whether this content would belong better in that document instead of this one. 🤔

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+Here is an example of usage of this tag:
+
+On the JSP context:
```
wincent, line 87: "On the JSP context" → "In the JSP page"

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+We couldn’t use both `require` and `use` in the same file. Because `use` property calls the modularization for alloy-ui modules and `require` follows ES6 modularized files.
+
+You can see more information about `aui:script` [here](https://help.liferay.com/hc/en-us/articles/360017882752-Loading-Modules-with-AUI-Script-in-Liferay-DXP).
+
+## `liferay-frontend:component` tag
+
+We can use this tag for removing JavaScript code from JSPs. This tag accepts a `module` property wether we can pass a relative path for the script to be loaded. Also, this tag registers automatically the component to [Liferay’s component global registry](https://github.com/liferay/liferay-frontend-guidelines/blob/master/dxp/liferay_component.md#register), depending on the `componentId` property. When using this tag, `namespace`(or `portletNamespace`) and `spritemap` props will turn available.
+
+Here is an example of usage of this tag:
```
wincent, line 85: I'd get rid of the ":" here. You have two sentences in a row ending with a colon.

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
jbalsas: No need

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+And when exporting the `myComponent.js` file we can define a default function there like:
+
+```js
+export default function myComponent(props) {
+	console.log(props.namespace);
+	// it will return the portlet namespace where the liferay-frontend:component tag were called
+	console.log(props.spritemap);
+	// it will return the path to the page's spritemap
```
wincent, line 106: Again, you can eliminate the extraneous detail here, and avoid a `console.log()` (which should never be in any code we write):(

```js
export default function myComponent({namespace, spritemap, ...props}) {
   // ... code that uses namespace, spritemap etc
}
```

(I added the `...props` there to show that we might get other stuff... I don't actually know what we could get passed here.)

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
jbalsas, line 19: Weird spacing around the ` combo for more details` link. It dos not seem to generate a valid url.

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
wincent: - "ES6" → just say "modern JS" (because otherwise you're making it sound like we can only use the already-ancient ES2015, which we've already moved far beyond, and we transpile a bunch of ES2019 stuff out-of-the-box in DXP).
- "We" → "we"

It's not so much "we don't have support for transpiling ES6 from taglibs" though, is it? It's more like "we don't transpile JS found in JSP files" — is that right?

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
jbalsas: I think this is a bit confusing because it's not clear how `aui:script` is related to ComboServlet...

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+By just placing JavaScript code inside an `aui:script` tag, you can take advantage of optimizing scripts and place them to be made available in the final compressed JavaScript bundle, see [combo for more details](https: / /github.com/liferay/liferay-frontend-guidelines/blob/45bad0334cb467e45e68aee322a2a4bc0b6979d3/dxp/javaScript_minification.md#comboservlet). It is worth mentioning that the same rule applies for the following use cases.
+
+#### Import an AUI/YUI module using `use` property
+
+When using the `use` property, the AUI / YUI module listed in the` use` list will be imported to the script data. If you want to know a little more about ScriptData, see this [link on how resources are injected into DXP](https://github.com/liferay/liferay-frontend-guidelines/blob/ea185686db8d562ac11cc2b95d2e9e7d7a0c6547/dxp/resource_injection.md#scriptdata).
+
+You can import several YUI / AUI modules within the same `use` property, as in this example:
+
+```jsp
+<aui:script use="liferay-upload">
```
jbalsas, line 28: It could be nice to show what this actually prints in our source code.

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+	console.log(props.namespace);
+	// it will return the portlet namespace where the liferay-frontend:component tag were called
+	console.log(props.spritemap);
+	// it will return the path to the page's spritemap
+}
+```
+
+This taglib can be very handy if you want to use modern ES6 features, since We don’t have support for transpiling ES6 from taglibs.
+
+Also, this tag adds some treatments for handling JavaScript on our JSPs like [adding special treatments for our built-in SPA framework](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-taglib/frontend-taglib/src/main/java/com/liferay/frontend/taglib/servlet/taglib/ComponentTag.java#L225).
```
wincent, line 112: - "treatments" → "treatment" (in both places)
- "on" → "in"

If you are going to link to a specific line, you should use an "evergreen" link (ie. one with a hash instead of a symbolic name like `master`); you can get that by pressing "y" on GitHub when you're looking at the page in question.

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
jbalsas, line 54: Same as for the other script, it would be good to show what this really prints to the page so it's easier to understand.

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
jbalsas: > Taking a look on the implementation, We aren't checking if JQuery is available on the page or not. Should I open a ticket for it?

No need. It simply means `A.$ = undefined`. Nothing bad happens for not checking since we do a simple assignment.

wincent (changes requested):
I left a lot of small comments inline so please forgive me.

forgiveness-please

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
jbalsas, line 77: We can use them in the same file, just not in the same `aui:script`

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+The purpose for the sandbox attribute is to wrap the contents of the script tag in a self-invoking function, and to enable short-handing the `$` and `_` variables rather than calling `AUI.$` and `AUI._`.
+
+#### Caveats:
+
+We couldn’t use both `require` and `use` in the same file. Because `use` property calls the modularization for alloy-ui modules and `require` follows ES6 modularized files.
+
+You can see more information about `aui:script` [here](https://help.liferay.com/hc/en-us/articles/360017882752-Loading-Modules-with-AUI-Script-in-Liferay-DXP).
+
+## `liferay-frontend:component` tag
```
jbalsas, line 81: Maybe let's split this to a different article? 
- dxp/aui_script.md
- dxp/liferay_frontend_component.md

I like demistifying as the title of the issue, but I don't think it's necessary for the articles. Just `aui:script` and `liferay-frontend:component` might be enough :D

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
+# Demystifying `aui:script` and `liferay-frontend:component` tags
```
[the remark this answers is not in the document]
jbalsas: Looks like Greg was adding all the feedback together while I was lazily pushing it out one by one.

See https://github.com/liferay/liferay-frontend-guidelines/pull/226/files#r499603480, but always pick what @wincent says before 😂

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: Sure, when I provided this example I was minding the reader will suppose it but We can reinforce :)

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: I'll update! Thanks for catching it up! XD

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: > I think you can drop the references to both $ and _. We do not want to encourage or even admit that jQuery and Underscore exist.

Makes sense omit this. I'll update

diegonvs: > I'd say absolutely not. This is a known issue, but we don't care: jQuery is off by default, and no code that we write from now on will ever rely on it.
> No need. It simply means A.$ = undefined. Nothing bad happens for not checking since we do a simple assignment.

Ok :)

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: > It's a tag, not a taglib (group of tags) ;)

On my mind tag and taglib means the same. I'll review whole doc :)

diegonvs: > "another" → "other" (or just remove the word entirely... less is more).

Ok mrs Wincent, thanks for the lang help here <3

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: > "have" → "has"

I replaced with: 

> We have three possible use cases when using this tag:

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: > Again here I'd drop "Pure" and say "Inline" (or "embedded")?

I got you thoughts here. I would prefer use `inline` since embedded can be interpret wrongly as a way to embed an external JS within the JSP body.

diegonvs: > I'd add an example to this use case

I'll add an example now

diegonvs: Does it looks good? I would go for a simpler and maybe practical use case that involves Java too

```jsp
<aui:script>
	Liferay.fire('myGlobalEvent', {
		'payload': <%= myDisplayContext.getPayload() %>
	});
</aui:script>
```

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: Makes sense replacing "ES6" occurrences with "modern JS" for improve the readability of non-javascript persons :p

diegonvs:
> Adding info about the defaultEventHandler API, what it is and how to use it to liferay-frontend:component may be useful too.

Hey @carloslancha, @jbalsas, @wincent, 

I don’t know too much regarding the defaultEventHandler API.
I just presumed the behaviour of this pattern as the way we found to pass data from a Java context to a JS context with a callback-like?

I definitely need some help here.

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: > I think this is a bit confusing because it's not clear how aui:script is related to ComboServlet...

Does it looks fine?

By just placing JavaScript code inside an `aui:script` tag, you can take advantage of optimizing scripts and place them to be made available in the final compressed JavaScript bundle. It can be achieved when appending the script to the [`AUI_SCRIPT_DATA` WebKey](https://github.com/liferay/liferay-portal/blob/67b569099146a4b999e2fad7d7d1a9794a337f0b/portal-kernel/src/com/liferay/portal/kernel/util/WebKeys.java#L55) and all these scripts will be applied at the [bottom of the page](https://github.com/liferay/liferay-portal/blob/e3ffac158e0ec5acc5c67069fbd7ba688d3c78d4/portal-web/docroot/html/common/themes/bottom.jsp#L44). Finally the [Aggregate Filter will take care](https://github.com/liferay/liferay-frontend-guidelines/blob/master/dxp/resource_serving.md#aggregate-filter), of creating the final bundle with all the JavaScript on the page. It is worth mentioning that the same rule applies for all the following use cases.

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: Done!

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: Done

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+On JSP context you may notice a huge number of `aui:script` and `liferay-frontend:component` tags.
```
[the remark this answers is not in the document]
diegonvs: done

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+## `aui:script` tag
+
+### What it does:
```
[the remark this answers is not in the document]
diegonvs: Updated

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+
+This taglib allow users to use and require another JS module/resources inside a JSP.
+
+This taglib have three possible use cases:
+
+1. Pure JS on the JSP
```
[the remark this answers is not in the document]
diegonvs: > "another" → "other" (or just remove the word entirely... less is more).
This tag allow users to use and require JS module/resources inside a JSP.

> "have" → "has"
We have three possible use cases when using this tag:

> The word "Pure" here might be a bit misleading. Can we just say "Inline JS inside the JSP"?
Done

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
```diff
@@ -0,0 +1,114 @@
...
+2. Import a AUI/YUI module
+3. Import a ES6 module
```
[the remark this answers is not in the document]
diegonvs: Great! I didn't knew it yet :p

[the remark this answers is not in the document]
diegonvs: > "a" → "an"

Import a modern JS module

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: > Again here I'd drop "Pure" and say "Inline" (or "embedded")?

> ("Pure" is kind of an implementation detail, as it hints at whether or not the JS is transpiled or mixed up with other stuff; in our case it is not transpiled but it's not "pure" either, because it can have JSP expressions, comments, declarations, directives, scriptlets etc interpolated inside it.)

I adopted "Inline" instead of "Pure"

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: > "using use" → "using the use"

✅

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: Yep, seems a bit lost. I'm working on tracking better as https://github.com/liferay/liferay-frontend-guidelines/pull/226#discussion_r499970785 comment :)

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
wincent: That works too.

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
wincent: I would even drop the `var` to make it brief (just inline it); eg. `Liferay.fire('some-event'`

The thing about changing that JSP expression from `<%= myDisplayContext.getPayload() %>` to `'myPayload'` is that you're blurring the lines between what the `<aui:script>` tag itself does, and how JSP itself works in general. This level of detail is extraneous. I know you want to be helpful, but by giving more information you're making your message less clear. This really could be as simple as:

> This example:
> 
> ```
> <aui:script>
>     // some JS code...
> </aui:script>
> ```
> 
> will be turned into:
> 
> ```js
> <script>
>     // some JS code...
> </script>
> ```

(Note: you should check what the output actually is. I don't think you get the IIFE without `sandbox`.)

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
wincent: To be honest, I don't think any non-JS people are going to be reading this. My main concern is that we don't use a term that refers to a standard we don't use any more.

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
wincent: That seems better, but I am not confident enough about my knowledge of the system to say whether it is right or not. @izaera: can you confirm that this is factually correct?

izaera: > Finally the Aggregate Filter will take care, of creating the final bundle with all the JavaScript on the page. It is worth mentioning that the same rule applies to all the following use cases.

I can't find any evidence of this part being true. In fact, it looks like [it's this guy](https://github.com/liferay/liferay-portal/blob/ebed38f329f4db6bfae067c46a26da3397597f2b/portal-impl/src/com/liferay/portlet/ScriptDataPortletFilter.java#L70) who writes it...

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: Yes, considering this to be injected via scriptData is completely wrong. AUI/YUI modules are loaded asynchronously. 

Replaced with:

_Each AUI / YUI module listed as a value for the `use` property will be [loaded asyncronously by the `AUI().use` function](https://github.com/yui/yui3/blob/master/src/yui/js/yui.js#L1323..L1331)._

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: WTF I wrote this!? I'll change

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: As used on https://github.com/liferay/liferay-frontend-guidelines/blob/master/dxp/resource_serving.md#javascript-bundle,

I just copied and paste the warning as the example below:


#### Import an AUI/YUI module using the `use` property 

> ⚠️ This is a deprecated use case, though it still works if someone configures it.⚠️


Also, I added a more detailed description on the beginning of the document as the following example:

## IMPORTANT

Infrastructure is sunsetting AlloyUI since when ["The status and direction of the frontend infrastructure in Liferay 7 & DXP"](https://liferay.dev/blogs/-/blogs/the-status-and-direction-of-the-frontend-infrastructure-in-liferay-7-dxp) blog post was released. Also, We are working hard in removing those components on [Remove AUI / YUI from Liferay Portal](https://issues.liferay.com/browse/LPS-98564) Epic. Please, avoid using AlloyUI JS components and tags, except for `aui:script` tag.

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: When loading an AUI component from the `frontend-js-aui-web` module, this component will turn available on the Liferay global object on the scope of the `aui:script`.

I would explain this. 👆

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: > Also, this link seems totally random and unrelated.

I would explain the user that this dependency must be declared on the package.json or provided by liferay-npm-bridge. Does make sense adding this info? Should I just remove it?

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: > Missing from this is an explanation of why you would want to use the as "operator", and guidance about what kind of alias to pick.

I was expecting the user to thought by the example but I can provide one more sentence for it

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: Yes, good insight w/ metal-dom example; I could use other example. Also, thanks for remembering for not providing very distracting examples. I'll update it!

diegonvs: I just created this example:

```jsp
<aui:script require="my-custom-lib/src/path/main as myLib">
	// ... code that use some function from myLib like `myLib.someFunction();`
</aui:script>
```

with this output example:

```js
Liferay.Loader.require('my-custom-lib/src/path/main', function(myCustomLibSrcPathMain) {
	try {
		(function() {
			var myLib = myCustomLibSrcPathMain;
			var $ = AUI.$;var _ = AUI._;
			// ... code that use some function from myLib like `myLib.someFunction();`
		})();
	} catch (err) {
		console.error(err);
	}
}
```

What do you think?

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: > #### Using `position` to control script placement

most clear. Thank you :)

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: Chema noticed the same error. I actually fixed with: 

> We couldn’t use both `require` and `use` in the same `aui:script` tag. Because `use` property calls the modularization for alloy-ui modules and `require` follows the [modern JS pattern](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import).

However your suggestion looks more clear. Thanks for helping <3

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: * "wether" → "whether", but I think you actually means something else? Like "in which" or "by which" or "through which" or "via which" (all of those work).

"through which" fits adequately. Thanks <3 

* "turn available" → "become available".

✅

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: I replaced this sentence

>Here is an example of usage of this tag:

> In the JSP page:

to

> For example, in the JSP page:

Does it looks well?

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: I'll update here. 

~~no-console even on docs~~
~~no-console even on docs~~
~~no-console even on docs~~
~~no-console even on docs~~
~~no-console even on docs~~
~~no-console even on docs~~
~~no-console even on docs~~

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: > If you are going to link to a specific line, you should use an "evergreen" link 

I asked myself about it when writing and I decided to adopt `master` as the branch to retrieve an updated doc. However, makes sense using "evergreen" for all the cases but it require more maintenance on docs for updating the branch and see if this info still valid.

I'll update it :p

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: comma killed 🔪

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: comma killed 🔪

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: Solved. The most complex english matter is prepositions. I try to memorize the preposition rules but It's complicated for latin-language speakers without too much daily practice 😢

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
diegonvs: I'll do once all the questions close

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
wincent: The problem with `master` links (or any link to any symbolic ref, like a branch name), is that they can break when files move. The worst class of links are those that link to a specific line, because lines can move around within a file. If a file moves, the user will get a 404 and think, "Oh, a busted link", but if lines move, the user will land on a page and think, "Huh? I'm not seeing anything like I expected to see", which is worse.

We discussed evergreen links in the past [here](https://github.com/liferay/liferay-frontend-guidelines/pull/201#discussion_r487035366), but I think it's time to document this, so I made:

https://github.com/liferay/liferay-frontend-guidelines/pull/228

dxp/demystifying_aui_script_and_liferay-frontend_component_tags.md
wincent: Yep.

diegonvs:
Ready for another review

karate-kid

dxp/aui_script.md
```diff
@@ -0,0 +1,148 @@
...
+				closeable: true,
+				message: 'I like drink tea while eating french potatoes',
+				type: 'success'
+			}
+		).render('#myId');
+	})()
+}
+```
+
+Note in the AlloyUI components, coming from the alloy-ui repository, classes are made available from the global variable `A` on the scope of the `aui:script`. Here as in this [code snippet the Menu component is defined](https://github.com/liferay/alloy-ui/blob/master/src/aui-menu/js/aui-menu.js#L39]).
```
izaera, line 77: I would change:

> on the scope of the `aui:script`. Here as in this

to:

> on the scope of the `aui:script` (as in this

If I understood the sentence correctly...

dxp/aui_script.md
```diff
@@ -0,0 +1,148 @@
...
+
+However, components made available from the `frontend-js-aui-web` module will be under the global Liferay variable. Like this [example here, `Liferay.Upload`](https://github.com/liferay/liferay-portal/blob/815f48f484351e18b61e4b9c9fbf40f0609bdc56/modules/apps/frontend-js/frontend-js-aui-web/src/main/resources/META-INF/resources/liferay/upload.js#L1541), used in the code example above.
+
+#### Import a ES module using `require` property
```
izaera, line 81: a ES module -> an ES module

dxp/aui_script.md
```diff
@@ -0,0 +1,148 @@
...
+
+Allows users to require a modern ES module as in the following example:
+
+Considering We already have `metal-clipboard` and `metal-dom` dependencies listed [on our module package.json](https://help.liferay.com/hc/en-us/articles/360018159771-The-Structure-of-OSGi-Bundles-Containing-npm-Packages-)
```
izaera, line 85: Considering We -> Considering we
listed on -> listed in ? @wincent ?

dxp/aui_script.md
```diff
@@ -0,0 +1,148 @@
...
+
+By default the name of the loaded resource will be the transformed path to camel-case default exported module/file:
```
izaera, line 87: -> By default, the name of the variable referring to the required resource will be composed by transforming the resource's path to camel case. For example:

dxp/aui_script.md
```diff
@@ -0,0 +1,148 @@
...
+
+```jsp
+<aui:script require="metal-clipboard/src/Clipboard">
+    // Note We have the provided path to require metal-clipboard/src/Clipboard normalized and camel-case as a variable name
```
izaera, line 91: -> // Note how the required `metal-clipboard/src/Clipboard` is made available in the `metalClipboardSrcClipboard` variable

dxp/aui_script.md
```diff
@@ -0,0 +1,148 @@
...
+	new metalClipboardSrcClipboard.default();
+</aui:script>
+```
+
```
izaera, line 95: I would add here:

> You can have a look at the details of the algorithm for composing variable names in the [`VariableUtil.generateVariable()` method](https://github.com/liferay/liferay-portal/blob/b69a6cc79c9cd2d3e13ef2879c6f158be0b1c094/portal-kernel/src/com/liferay/portal/kernel/servlet/taglib/aui/VariableUtil.java#L33).

dxp/aui_script.md
```diff
@@ -0,0 +1,148 @@
...
+Also, We can set an alias with the `as` word after the provided path. It will be look like:
```
izaera, line 96: We -> we


It will be look like -> It would look like  (sounds better to me :shrug: )

dxp/aui_script.md
```diff
@@ -0,0 +1,148 @@
...
+
+```jsp
+<aui:script require="my-custom-lib/src/path/main as myLib">
+	// ... code that use some function from myLib like `myLib.someFunction();`
```
izaera, line 100: that use -> that uses

dxp/aui_script.md
```diff
@@ -0,0 +1,148 @@
...
+
+The output will be:
+
+```js
+Liferay.Loader.require('my-custom-lib/src/path/main', function(myCustomLibSrcPathMain) {
+	try {
+		(function() {
+			var myLib = myCustomLibSrcPathMain;
+			var $ = AUI.$;var _ = AUI._;
+			// ... code that use some function from myLib like `myLib.someFunction();`
```
izaera, line 112: -> that uses

dxp/aui_script.md
```diff
@@ -0,0 +1,148 @@
...
+		console.error(err);
+	}
+}
+```
+
+You can find more information on [this link](https://help.liferay.com/hc/en-us/articles/360017882752-Loading-Modules-with-AUI-Script-in-Liferay-DXP#loading-es2015-and-metaljs-modules)
+
+#### Using `position` to control script placement
+
+This property will change the way in which / how the script will be injected into the page. By default, those scripts that are wrapped by `aui:script` will be added to the minified files as described [here](#inline-js-on-the-jsp).
```
izaera, line 124: I would remove `/ how`

dxp/aui_script.md
```diff
@@ -0,0 +1,148 @@
...
+</aui:script>
+
+<aui:script sandbox="<%= true %>>
+   var myVar = 'foo'; // This variable will NOT be available globally
+</aui:script>
+```
+
+#### Caveats
+
+You can't use both require and use in the same tag. use is for declaring dependencies on AUI modules, while require gives access to ES modules.
```
izaera, line 146: require -> `require`
use -> `use`

Note that each term appears twice ;-).

dxp/aui_script.md
```diff
@@ -0,0 +1,148 @@
...
+
+You can see more information about `aui:script` [here](https://help.liferay.com/hc/en-us/articles/360017882752-Loading-Modules-with-AUI-Script-in-Liferay-DXP).
```
izaera, line 148: Just an accessibility recommendation that @marcoscv taught us: 

Instead of linking the word `here` (which makes the user having to visit the link to know what it is), try to describe in the link's text what you are exactly linking. For example, in this case, you could change `here` to `in this article`, or something akin.

dxp/liferay_frontend_component.md
```diff
@@ -0,0 +1,28 @@
+# Demystifying `liferay-frontend:component` tag
+
+We can use this tag for removing JavaScript code from JSPs. This tag accepts a `module` property through which We can pass a relative path for the script to be loaded. Also, this tag registers automatically the component to [Liferay’s component global registry](https://github.com/liferay/liferay-frontend-guidelines/blob/bc6dae8514af04a6384a7bea9d4ddf266087ffc5/dxp/liferay_component.md#register), depending on the `componentId` property. When using this tag, `namespace`(or `portletNamespace`) and `spritemap` props will become available.
```
izaera, line 3: 1. We -> we

2. registers automatically the component to -> registers automatically the component with

Regarding 2, I think that's the correct form (see https://idioms.thefreedictionary.com/register+with), however I would like to invoke @wincent 's wildcard here :sweat_smile:

dxp/liferay_frontend_component.md
```diff
@@ -0,0 +1,28 @@
...
+```jsp
+...
+<liferay-frontend:component
+	componentId="myComponent"
+	module="path/to/my/js/myComponent.js"
+/>
+...
+```
+
+And when exporting the `myComponent.js` file We can define a default function there like:
```
izaera, line 16: We -> we

dxp/liferay_frontend_component.md
```diff
@@ -0,0 +1,28 @@
...
+
+```js
+export default function myComponent({namespace, spritemap, ...props}) {
+	// ... code that uses namespace, spritemap etc
+}
+```
+
+This tag can be very handy if you want to use modern JS features, since We don’t have support for transpiling modern JS from taglibs.
```
izaera, line 24: We -> we

izaera:
Very interesting content :+1:

dxp/aui_script.md
wincent: Yeah, "listed in".

dxp/aui_script.md
wincent: Even better than "in this article" is including the name of the article, because if the user asks the screenreader for a list of links on the page, "in this article" is only a small improvement over "here", but the full title is very informative indeed:

> You can see more information about `aui:script` in ["Loading Modules with AUI Script in Liferay DXP"](https://help.liferay.com/hc/en-us/articles/360017882752-Loading-Modules-with-AUI-Script-in-Liferay-DXP).

dxp/liferay_frontend_component.md
wincent: Yes, "with" is better than "to" in this case, and you can additionally make the flow sound more naturally by changing the order of the words:

eg. "registers automatically the component with" → "automatically registers the component with"

dxp/aui_script.md
diegonvs: Makes sense :) Updating

dxp/aui_script.md
diegonvs: ✅

dxp/aui_script.md
diegonvs: ✅

dxp/aui_script.md
diegonvs: ✅

dxp/aui_script.md
diegonvs: ✅

dxp/aui_script.md
diegonvs: I searched for this piece of code somewhere but I didn't found. Thank you! :)

dxp/aui_script.md
diegonvs: english

dxp/aui_script.md
diegonvs: ✅

dxp/aui_script.md
diegonvs: ✅

dxp/aui_script.md
diegonvs: I was probably thinking in portuguese when writing this part

dxp/aui_script.md
diegonvs: Note that each term appears twice ;-).

Thanks for point up! :)

dxp/aui_script.md
diegonvs: Thanks for the suggestions. I remember using this pattern somewhere but I found other using just `here`. Not on our repo but somewhere. I will use this pattern from now everywhere :)

diegonvs: Replaced every [here] match with a more descriptive text :)

dxp/liferay_frontend_component.md
diegonvs: Thanks for the feedback <3

dxp/liferay_frontend_component.md
diegonvs: ✅

dxp/liferay_frontend_component.md
diegonvs: ✅

diegonvs:
Updated! c'mon

karate-kid

I think this doc is now looking perfect

dxp/aui_script.md
izaera: Yes, much better what @wincent says :+1:

dxp/aui_script.md
izaera: :+1:

dxp/aui_script.md
izaera: Yeah, sounded more like Spanish than English :sweat_smile:

wincent (approved):
Screenshot 2020-10-14 -092657-5vkObqTX@2x

As much as I would like to keep this splendid conversation going for a bit longer, all good things must come to an end. As a final sanity check I looked at the GitHub rendered output to see if it looks clean and correct (it does), so I think this should go in now and if there are any problems with it we can follow-up with fixes. The most important thing is the "sunsetting" disclaimer at the top, so I think we're good here.

